PR #26, opened 2018-09-07
[OCDM] Rewrite some old-school container iteration code.

charlie-ht:


Source/ocdm/open_cdm.cpp
```diff
@@ -173,6 +173,18 @@ class AccessorOCDM : public OCDM::IAccessorOCDM {
         ::OCDM::ISession::KeyStatus Status () const {
             return (_status);
         }
+        const char* StatusString() const {
```
pwielders: Like WebKit, WPEFramework has a basic library, found in core, that has templates to map enum values to strings. This template allows for optimizations, so if enums need to be printed suggest to use the template Core::EnumerateType<enumType> etc..

Source/ocdm/open_cdm.cpp
```diff
@@ -285,44 +298,31 @@ class AccessorOCDM : public OCDM::IAccessorOCDM {
...
-                index  = session->second.begin();
-
-                while ((index != session->second.end()) && (*index != paramKey)) { index++; }
-
-                if (index != session->second.end()) {
-                    container = &(session->second);
-                }
-                else {
-                    session++;
+            // FIXME: We shouldn't be looping through every available session really,
```
pwielders: I think the root cause of this lies in a requirement enforced by WebKit usage. There was "no" way to link the EME session created on one hand in the WebKit to the gstreamer actually using it. The only linking pin between the EME session and the gstreamer was the KeyId. Hence why we *must* loop through all the session if we want to find the keyId here that is offered by the gstreamer. So unless WebKit forwards the SessionId to the gstreamer, so the gstreamer can pick up the session by the sessionId, I see no solution except for iterating over all the sessions untill we find the KeyId.

Source/ocdm/open_cdm.cpp
```diff
@@ -285,44 +298,31 @@ class AccessorOCDM : public OCDM::IAccessorOCDM {
...
-                result = (index->Status() == status);
-            }
-
-            if (result == false) {
+            if (!result) {
                 _interested++;
 
                 _adminLock.Unlock();
 
-                if ((container != nullptr) && (index != container->end())) {
```
pwielders: Since you are now looping trough all entries, even if onw is found, we do nolonger have this diagnostic information again, which is a pitty as these trace lines helped in debugging the OCDM functionality. I do think these trace lines where quite usefull.

Source/ocdm/open_cdm.cpp
```diff
@@ -285,44 +298,31 @@ class AccessorOCDM : public OCDM::IAccessorOCDM {
...
-                    TRACE_L1("Waiting for KeyId: %s, current: %d", paramKey.ToString().c_str(), index->Status());
-                }
-                else {
-                    TRACE_L1("Waiting for KeyId: %s, current: <Not Found>", paramKey.ToString().c_str());
-                }
-
                 uint64_t now (Core::Time::Now().Ticks());
 
-                if (now < timeOut) {
+                if (now < timeOut)
```
pwielders: Minor: For maintainability we have a coding rule to always scope single lines.

Source/ocdm/open_cdm.cpp
charlie-ht: I've not come across that, but sounds very handy, I will take a look!

Source/ocdm/open_cdm.cpp
charlie-ht: I realise it's a short-coming, I placed the comment here so future readers don't have to wonder why we're being piggy. Basically a note for future improvement if possible.

Source/ocdm/open_cdm.cpp
```diff
@@ -285,44 +298,31 @@ class AccessorOCDM : public OCDM::IAccessorOCDM {
...
-
-                if (index != session->second.end()) {
-                    container = &(session->second);
-                }
-                else {
-                    session++;
+            // FIXME: We shouldn't be looping through every available session really,
+            // the user of decryption should have a session handle, since sessions are supposed
+            // to be private by the spec.
+            for (const auto& sessionKeyPair : _sessionKeys) {
```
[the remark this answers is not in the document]
charlie-ht: Well this is subjective, I can make the code skip the extra checks by just `break`ing after setting result=true (that should have been done anyway). But remember having more than 2 sessions with more than 4 keys each is essentially unheard of in practice, so I would challenge you to measure a performance degradation that is above a noise ceiling here.

About auto, the most up-to-date guidelines I know of are the ones spearheaded by Bjane ("The C++ Core Guidelines"). Specifically the refactoring fixes guidelines ES.1 "Prefer the standard library to other libraries and to “handcrafted code”", P.1: "Express ideas directly in code", P.3: "Express intent" and ES.11: "Use auto to avoid redundant repetition of type names". The core guidlines are browsable here https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines. That's why I do it these ways. I agree auto can be overused and obfuscate the type names, but I don't see the problem in this example.

Source/ocdm/open_cdm.cpp
charlie-ht: They are still being traced, the looping through all entries is an easy fix.

Source/ocdm/open_cdm.cpp
charlie-ht: Noted, I'm so used to having to delete them from WebKit, I prefer to have them as well :-)

PR #28, opened 2018-09-12
JSON: Validator functionality added

HaseenaSainul:


pwielders:
Quick review done. Looks good (compact) 
One question:
I have the impression that the validator needs to be run separate from the initial deserialize. Was hoping we could integrate it with the deserialization, meaning we only have to parse it once, and during the parse, check the fields. This means less performance impact and checking and paring a field lies closer together.
Then If you find an issue (invalid json) note it in the object and after parsing (FromString) to the object, you can ask the object if errors where encountered, if so a warning can be issued.
This functionality needs to be used by each plugin as well. Hope we can simplify the interface by just "asking" after the FromString, if the JSON was valid and if not raise a warning.

HaseenaSainul:
Thank you for the comments :-)
In the current implementation, while loading the Metrological WPE Frame UI, it is sending the JSON string value (configuration) of all plugins to the main.js  (While requesting the plugin info from the main.js api.getControllerPlugins() : https://github.com/WebPlatformForEmbedded/WPEFrameworkUI/blob/master/build/js/main.js#L353. It will simply return all the configuration files which is stored during the plugin loading time: https://github.com/WebPlatformForEmbedded/WPEFramework/blob/master/Source/WPEFramework/Controller.cpp#L224). If any format error in any of the JSON file, the json.parse in the main.js will fail. 

Also the deserialisation of the configuration data of each plugin is happening only after activating the plugin (in a later time and based on demand). So If we do the validation against each plugin during the utilisation time, we will get the issue in later time only.

woutermeek:
@HaseenaSainul we are going to solve this in a different way, because the JSON objects have been completely refactored.

PR #32, opened 2018-10-16
Aamp integration changes

HaseenaSainul:


pwielders:
The configure is already available in the interface, so a bit lost why it is added. 
The Terminat should *NOT* be called externally as it is part of the liftetime management, it will automatically be called if you release the interface. However for this to work (Life time managament, not the terminate) you need to have a patch (incorrect lifetime management currently) so this pull-request is not needed.

HaseenaSainul:
Closing this PR, as per the discussion with @pwielders , the player platform implementation in archive is more good than the shared library.

PR #38, opened 2018-11-23
[Netflix] FactoryReset and Visibility APIs are added

HaseenaSainul:


Source/interfaces/ProxyStubs.cpp
```diff
@@ -2132,21 +2146,35 @@ namespace ProxyStubs {
             return reader.Text();
         }
 
-        virtual void SystemCommand(const string& command)
```
wouterlucas: Will this break the systemcommand API for NF4?

HaseenaSainul: No it will not, just rearranged after factory reset api. Factory reset and visibility apis are newly added for NF5, but once we merge NF5 changes in WPEPluginNetflix it will not make any issue for NF4 since added NF5 Netflix apis under macros

PR #39, opened 2018-12-06
[VirtualInput] Timed release and simultaneous key press release handl…

aswaunni:
Timed RELEASE in case a RELEASE key is not followed after a PRESSED key for a given time and handling simultaneous key press (second press is given only after releasing the first key).

### DIFF
--- a/Source/plugins/VirtualInput.cpp
+++ b/Source/plugins/VirtualInput.cpp
@@ -289,6 +289,8 @@ ENUM_CONVERSION_END(PluginHost::VirtualInput::KeyMap::modifier)
                     sendModifiers = element->Modifiers;
                 }
 
+                if ((!pressed) && (_pressedCode != code))
+                    result = Core::ERROR_ALREADY_RELEASED;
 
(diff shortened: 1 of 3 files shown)


Source/plugins/VirtualInput.cpp
```diff
@@ -322,6 +333,7 @@ ENUM_CONVERSION_END(PluginHost::VirtualInput::KeyMap::modifier)
         void VirtualInput::RepeatKey(const uint32_t code)
         {
             AdministerAndSendKey(REPEAT, code);
+            KeyEvent(false, _pressedCode, _keyTable);
```
pwielders: I would not expect this here. How does it work Now it looks like once 1 repeat key event is send, the key is forced to release. I would expect this to happen once N number of repeatkeys where send. Does the repeat key feature still work with this patch?

aswaunni: I'll add the change for release after N repeatkeys. But from where N will get the value, should that be passed as config value.

PR #46, opened 2019-01-28
Bram/cmake refactor

bramoosterhuis:


MFransen69:
Hi Bram,
My CMake knowledge is still limited ;) but does this mean that (at the moment) there is no difference to release/Minsize release/Release with debug info? Should for example not additional flags be set like "-g0" for debug or is that done by CMake when DEBUG is set?

Best regards, Marcel

Source/provisionproxy/CMakeLists.txt
```diff
@@ -14,8 +14,9 @@ set(PUBLIC_HEADERS
         )
 
 target_link_libraries(${TARGET}
-        PRIVATE
+        PUBLIC
           ${NAMESPACE}Core::${NAMESPACE}Core
+        PRIVATE
```
pwielders: I think you lost me again :-) I see: find_package(libprovision REQUIRED) oke so I thought the rule was that if we need to include it in a build, i would just add libprovision::libprovision like all the other packages, but now all off a sudden metrological::libprovision...... How can I now I have to include metrological here ?
This is in line with cmake :-) But needs clarification.

pwielders: Also saw Find OpenSSL in this package but nowhere it is included and I think it can go. Right ?

PR #48, opened 2019-02-05
[Process] WaitProcessCompleted return value corrected

HaseenaSainul:


Source/core/Process.h
```diff
@@ -601,6 +601,10 @@ namespace Core {
                 }
             }
 
+            if (IsActive() == false) {
```
pwielders: I think this is violating a coding rule, Only one return (although there are exceptions, with good reasons) but here I suggest 1 return:
return (IsActive() == false ? Core::ERROR_NONE : Core::ERROR_TIMEDOUT)
and than there is only 1 return :-)

HaseenaSainul: Updated the return statements

PR #51, opened 2019-02-09
[WPE-221][INTERFACES] Add SystemCommands iface.

pawel-stanek-ml:
A very simple one; with just one command for the moment: USBReset.


pawel-stanek-ml:
This doesn't require rebase.

PR #52, opened 2019-02-11
[Process] PGID of forked child is changed to new group

HaseenaSainul:


HaseenaSainul:
The logic for handling child process are moved to launcher, hence closing this PR

PR #53, opened 2019-02-13
[websocket] Resolved issue related to file server-client.

aswaunni:


Source/websocket/WebTransfer.h
```diff
@@ -308,7 +308,7 @@ namespace Web {
                         _state = TRANSFER_UPLOAD;
                         _request.Verb = Web::Request::HTTP_PUT;
                         _request.Path = '/' + destination.Path().Value().Text();
-                        _request.Body(Core::ProxyType<FILEBODY>(&_fileBody));
+                        _request.Body(Core::ProxyType<FILEBODY>(_fileBody));
```
pwielders: Indeed in the area of the ProxyType a lot has changed and I think now it could "automagically" transfer from the specialized to the base type. Could you try compiling it also wothout the "Core::ProxyType<FILEBODY>" and just write: _request.Body(_fileBody); This way we leave it up to the compiler to properly convert/cast it.
If you run into issues let me know.

aswaunni: I'll try that

aswaunni: Its not compilable, since _fileBody is  ProxyObject and Body() expects ProxyType.

pwielders: Ohh indeed. Than it cannot be automagically converted. Thanks.

PR #55, opened 2019-02-14
[FRAMEWORK] IDS management improvement

pawel-stanek-ml:
[FRAMEWORK] IDS management improvement


pawel-stanek-ml:
As discussed with Pierre offline IDs < 0x40 should not be listed and the header mustn't be in the core.

Source/com/com.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../core/Ids.h"
```
pwielders: s discussed. Ids have no meaning in the context of Core, so this Id file needs to be in COM (in this directory). The IUnknown should be moved to here as well. The ids file in this directory should then contain only the IDS < 0x40, as this range is reserved for the WPEFramework.
The actual interfaces directory should then hold another Ids.h Which should then hold the ids >= 0x40

Source/core/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ set(PUBLIC_HEADERS
         FileSystem.h
         Frame.h
         IAction.h
+        Ids.h
```
pwielders: Move to COM

Source/core/Ids.h
```diff
@@ -0,0 +1,100 @@
...
+    ID_STATECONTROL = 0x0000002A,
+    ID_STATECONTROL_NOTIFICATION = 0x0000002B,
+    ID_SUBSYSTEM = 0x0000002C,
+    ID_SUBSYSTEM_NOTIFICATION = 0x0000002D,
+    ID_SUBSYSTEM_INTERNET = 0x0000002E,
+    ID_SUBSYSTEM_LOCATION = 0x0000002F,
+    ID_SUBSYSTEM_IDENTIFIER = 0x00000030,
+    ID_SUBSYSTEM_TIME = 0x00000031,
+
+    ID_BROWSER = 0x00000040,
```
pwielders: This file should be split in WPEFramework specific IDS < 0x40 in COM and IDS >= 0x40 in interfaces.

Source/core/Ids.h
```diff
@@ -0,0 +1,100 @@
+#pragma once
+
```
pwielders: This needs to go in a namespace. All our files are in WPEFramework namespace. For the IDS < 0x40 I suggest the namespace as deplued by the COM module. The IDS >= 0x40 should be in the Exchange namespace.
By definition not in the global namespace.

Source/core/Portability.h
```diff
@@ -567,7 +568,7 @@ namespace Core {
 	};
 
     struct EXTERNAL IUnknown {
```
pwielders: MOve this into the COM module.

pawel-stanek-ml: Unfortunately this requires more thinking since CORE uses IUnknown (Service.h) and it's a separate think so should not block this so I suggest to do it separately.

pawel-stanek-ml:
For now  I went for the ifaces IDs file (<0x40) located in the interfaces folder using WPEFramework Exchange namespaces.
The COM one will be possible to do after moving IUnknown to the COM which doesn't look easy doable by the quick glance due to CORE using it as noted in another comment, so let's do it in a separate step(s)

Source/core/Portability.h
pwielders: Well lets, here than not use the ID as found in the enum lets leave this hardcode, so nu use for the Ids.h. It is a WPEFramework ID anyway and move the Ids.h to the COM layer.

pawel-stanek-ml: Ok. I'll add COM Ids too and leave IUnknown behind using hardcoded value.

pwielders: Great, Thanks

pawel-stanek-ml:
We have Interfaces and COM Ids used all over the place. Only IUnknown is left out

PR #64, opened 2019-03-01
[OCDM] Recreate RPC::CommunicatorClient connection of OCDMAccessor wh…

Damian-Danylko:
…en server is interrupted by crash or deactivation

Solution for WPE-252, WPE-256


Damian-Danylko:
Fix will be provide on the separate branch.

PR #74, opened 2019-03-15
[Proxy] Cleanup sequence changed to avoid crash in the WPEFramework d…

HaseenaSainul:
…ue to crash in the WPEProcess


HaseenaSainul:
Closing this issue, since this is already covered in another fix

PR #75, opened 2019-03-18
[JSONRPC] Allow to register a "void" JSONRPC handler.

pawel-stanek-ml:


pawel-stanek-ml:
Note the changes are convenience ones: An API client can explicitly say: "I'm not interested in input parameters. Do not pass them to me" or  "Gimme MyParamsJsonContainerStructure instead of passing me a string I'd need to deserialize in method". Moreover lambdas work with it and OUTBOUND can also be a string

pwielders:
The idea is good to have the posibility to have a void parameter. But I assume that you could then also have a void response. So suggest to have these methods for both template parameters.
I was wondering why the "static" method was needed, I think you can more easilly solve this using compiletime diffrentiation.
See my implementation, you can now use "void" for INBOUND or OUTBOUND.

pawel-stanek-ml:
I don't think void on OUTBOUND is really an option since there is always some response: it's either error or success message.
The static method was so you could not only use JSON::String but also std::string as outbound.
What do you mean by compile time diffrentiation? This will solve compile time as well.

pawel-stanek-ml:
Could you give me some pointers at the implementation?

pawel-stanek-ml:
> Could you give me some pointers at the implementation?

Found it. Taking a look

pwielders:
OUTBOUND would really be an option, maybe you have not run into it but:
Getter only: Will have no INBOUND, will have an OUTBOUND
Setter only: Will have INBOUND, will have no OUTBOUND. (return value is sufficient to report a succefull set)
A notification JSONRPC method, do not think requires OUTBOUND, it might just be a fire and forget method

pawel-stanek-ml:
I ran into it and this is why I came up with my solution and at first made it optional too but but looked at the PluginHost::Invoke and say in case ERROR_NONE the Result member is set to OUTBOUND's value. In case of SUCCESS and specifying OUTOUND as void I take it'll be there but an empty string. Is that ok because we came to the conclusion it's be good to have something there and sice we didn;t like "0" (0 number as string we decided to go with "OK" string as it was in case of Restfull API so in pracice the OUTBOUND was always used.
So is empty Result in the response OK?

pawel-stanek-ml:
If it is OK I guess it means we should skip result description in the spec for setters/notifications.

pwielders:
Yip, agree, if the error code is '0' but there is nothing return, like in RemoteControl.1.delete lets make sure that the result is set to null, and pray it does not crash :-) If it crashes, we need to fix it :-)
According to the sepc: https://www.jsonrpc.org/specification there must be a result, so suggest to return: null. However, would be fine with an empty string as well.

PR #78, opened 2019-03-21
[OCDM] Adapter gstreamer version compile fixes

pawel-stanek-ml:


Source/ocdm/adapter/gstreamer/open_cdm_adapter.cpp
```diff
@@ -4,6 +4,21 @@
 #include <gst/gst.h>
 #include <gst/base/gstbytereader.h>
 
+inline bool mappedBuffer(GstBuffer *buffer, bool writable, uint8_t **data, uint32_t *size)
```
pawel-stanek-ml: Note this is copy&paste from SVP version but I came to the conclusion it's not worth creating a separate header and source files  just for this.

modeveci:
It is ok, my bad I could not test on default gstreamer impl. Thanks

Source/ocdm/adapter/gstreamer/open_cdm_adapter.cpp
modeveci: It is ok, it is already an inline func.

pawel-stanek-ml:
No problem at all.
Will you merge it or I should?

PR #84, opened 2019-03-27
Synchronize config.json location (C++ & cmake)

dwrobel:


bramoosterhuis:
I think it better to move the definition to this target_compile_definitions here. https://github.com/WebPlatformForEmbedded/WPEFramework/blob/577ac1443a6628eb02ed7671ecbda25a7eee8a62/Source/WPEFramework/CMakeLists.txt#L12

Maybe we also want to be able to set it from the build system? Then introduce a variable and default it to the current path. Then use that variable to install the config and set the define.  Like this: https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/blob/master/Compositor/CMakeLists.txt#L16

pwielders:
@dwrobel @bramoosterhuis 
Have we concluded this. Can we merge it?

dwrobel:
@pwielders I'm working on that...

pwielders:
Oke. Just send me a trigger if it is complete. Thanks.

dwrobel:
This is ready for further review, but even if you would be fine with it, please don't merge until I verify it also on EOS _(I'm using it so far on PC for local build))_.

>definition to this target_compile_definitions here. 

Added. However, as GenericConfig.cmake is included in the end of `CMakeLists.txt` I couldn't move it to the `CMakeLists.txt`.

>Maybe we also want to be able to set it from the build system?

It's already there _(see: CONFIG_INSTALL_PATH)_, but no one used it so far. So, I just remove the hardcoded `etc` part with `CMAKE_INSTALL_SYSCONFDIR` from GNUInstallDirs.

dwrobel:
>please don't merge until I verify it also on EOS

Tested also on EOS, works fine, so It's ready for review.

pwielders:
@bramoosterhuis can you review this and if it is oke, merge it?

bramoosterhuis:
Perfect, nice to intoduce GNUInstallDirs. Now I can remove that from the do-do list.

PR #92, opened 2019-04-05
Do not use aliased argument for 'restrict' parameter

dwrobel:
Fixes:
 - WPEFramework/Source/core/ProcessInfo.cpp:55:43: warning: passing argument 2 to restrict-qualified parameter aliases with argument 1 [-Wrestrict]
         int32_t length = readlink(buffer, buffer, maxLength - 1);
                                  ~~~~~~  ^~~~~~
 - protect from accessing 'buffer' memory beyond 'maxLength',

 - use /proc/*/comm instead of /proc/*/status as it is easier to parse
   and it doesn't involve sscanf() usage,

 - always append '\0' to the returned process name.

readlink(2) documentation: http://pubs.opengroup.org/onlinepubs/9699919799/functions/readlink.html

### DIFF
--- a/Source/core/ProcessInfo.cpp
+++ b/Source/core/ProcessInfo.cpp
@@ -51,26 +51,36 @@ namespace Core {
 #else
     void ProcessName(const uint32_t pid, TCHAR buffer[], const uint32_t maxLength)
     {
-        snprintf(buffer, maxLength, "/proc/%d/exe", pid);
-        int32_t length = readlink(buffer, buffer, maxLength - 1);
+        ASSERT(maxLength > 0);
 
-        if (length > 0) {
-            buffer[length] = '\0';
-        } else {
-            int fd;
+        if (maxLength > 0) {
+            char procpath[48];
+            snprintf(procpath, sizeof(procpath), "/proc/%u/exe", pid);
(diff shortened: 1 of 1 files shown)


Source/core/ProcessInfo.cpp
```diff
@@ -51,19 +51,30 @@ namespace Core {
 #else
     void ProcessName(const uint32_t pid, TCHAR buffer[], const uint32_t maxLength)
     {
-        snprintf(buffer, maxLength, "/proc/%d/exe", pid);
-        int32_t length = readlink(buffer, buffer, maxLength - 1);
+        assert(maxLength > 0);
```
pwielders: please use ASSERT i.s.o assert

Source/core/ProcessInfo.cpp
```diff
@@ -51,19 +51,30 @@ namespace Core {
...
+
+        if (maxLength == 0)
+            return;
```
pwielders: Only one exit point allowed in our code. Maintainability issue if havinf multiple returns

dwrobel: Yeah, I gotcha.

PR #96, opened 2019-04-10
Avoid polutting global namespace with `OK` define.

dwrobel:
Fixes clash with playready:

sysroots/dcx960-debug/usr/include/playready3x/bignum.h: In function 'DRM_BOOL PK::multiply(const digit_t*, DRM_DWORD, const digit_t*, DRM_DWORD, digit_t*)':
sysroots/dcx960-debug/usr/include/playready3x/bignum.h:565:21: error: lvalue required as left operand of assignment
         DRM_BOOL OK = TRUE;                                             \
                     ^

### DIFF
--- a/Source/core/Portability.h
+++ b/Source/core/Portability.h
@@ -371,9 +371,7 @@ typedef DEPRECATED signed long long sint64;
 #define TRUE (!FALSE)
 #endif
 
-#ifndef OK
-#define OK (0)
-#endif
+namespace WPEFramework { constexpr uint32_t OK = 0; }
 
 #ifdef __WIN32__
 #define SYSTEM_SYNC_HANDLE HANDLE


Source/core/Portability.h
```diff
@@ -371,9 +371,7 @@ typedef DEPRECATED signed long long sint64;
 #define TRUE (!FALSE)
 #endif
 
-#ifndef OK
```
pwielders: Lets keep the preprocess check around it. I would not like to add the OK by default if someone (wherever it is coming from) already defined it. Also do not know if it is actually used in the WPEFramework code.

dwrobel: Someone can use:
```
#define OK "OK"
```
I'm not sure you're interested in `ifdef'ing` around it.

pwielders: at the end the result (and thus the compiler error) will be the same as the preprocess runs prior to compilation it will replace then all our OK definitios in the code with "OK" before the compilation starts.
I think (but have not tested it) that at the end the OK is not used anymore (i ligned up some changes where the OK -> Core::ERROR_NONE as the where returning the OK as an error code. Legacy code I think. However replacing there the OK with ERROR_NONE probably means that the OK is not used anymore anyway. Maybe we could remove it, or lets add the deprecated label to the OK definition, so at least we get warnings if someone is using it.

dwrobel: >I think (but have not tested it) that at the end the OK is not used anymore 

So, should I still wrap it with `#ifndef` or it would be no longer necessary?

pwielders: I will remove the OK completely from the code and see where the compile stops :-)

PR #98, opened 2019-04-12
JsonObject/JsonValue improvements

mrcomcast123:
My changes should not impact anybody not using Variant/VariantContainer.

I made several changes to Variant and VariantContainer so that we can support full json with arrays and nested json objects.

I moved Variant/VariantContainer down in the JSON.h file so that ArrayType is declared first because Variant can now hold an ArrayType inside it (as well as a VariantContainer inside it).

I made the following changes that touch code outside those two classes but I don't think it would impact anybody:

core/JSON.h: made Container::ElementIterator and ArrayType:ElementIterator public so that Variant can call it.

core/JSON.h: added new method ArrayType::Get(const uint32_t index) which works the same as index operator but looks cleaner if ArrayType is a pointer. parray->Get(i) vs (*parray)[i].

jsonrpc/Channel.h: updated Subscribe methods to construct inbound from string, so that my handler can take a const reference like myHandler(const JsonObject& json).  It wouldn't compile otherwise.

plugins/JSONRPC.h: added another JSONRPC::Register method which defaults the template params to VariantContainer so that in my plugin i can just do this:  Register("foo", &Thing::foo, this);

I have built out our first wpe plugin replacement for one of our previous services and integrated that into our main application.  This work requires these changes.

Thanks

Mark Rollins


pwielders:
Hi Mark,

Must admit, it is a lot and since it seems to be only additions and thus not influencing any other JSON Objects on the Variants, happy to merge it (more functinality is always good :-) ). However before doing so I want to make sure it does not break any JSONRPC functionality already implemented. We are in the process of implementing the JSON RPC interfaces on all plugins and I do not want to interrupt this work.
Please give me a few days to validate that is pull request does not break it. If it does not break teh current build I will merge it to the maste.

BR Pierre

PR #100, opened 2019-04-15
[JSON] Serialise empty unquoted strings to null

sebaszm:
Make sure we output valid JSON on empty strings.

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -1113,12 +1113,14 @@ namespace Core {
                     offset++;
                 }
                 if (result < maxLength) {
+                    const std::string& source = (((_value.empty() == true) && (UseQuotes() == false))? "null" : _value);
+
 #ifdef __WIN32__
 #pragma warning(disable : 4996)
 #endif
 
                     // Write the amount we possibly can..
-                    uint16_t written = static_cast<uint16_t>(_value.copy(&(stream[result]), maxLength - result, offset - 1));
+                    uint16_t written = static_cast<uint16_t>(source.copy(&(stream[result]), maxLength - result, offset - 1));
 
 #ifdef __WIN32__
 #pragma warning(default : 4996)


Source/core/JSON.h
```diff
@@ -1113,12 +1113,14 @@ namespace Core {
                     offset++;
                 }
                 if (result < maxLength) {
+                    const std::string& source = (((_value.empty() == true) && (UseQuotes() == false))? "null" : _value);
```
pwielders: Do not put std:: in fromt of the string. There is a possibility (never tested) that string -> std::string or std::wstring) and for that "null" -> _T("null") Also is it not a bit tricky to make this a refernce to a std::string ?

MFransen69: Please use _T("null") for string literals (for if we ever want to go multibyte) and remove the std:: before string.

Assigning a temporary to a const& is allowed here (the "null" will become a string of course) but dependent on the optimizer the string("null") will always be created or only if the true branch of the ? is taken (I checked the VS2017 optimizer and it will only create the string("null") in case the ? true branch is chosen btw). But independent of this it would be better I think to prevent the sting creation from the literal completely and the new/delete that come with it (as this code will be called quite a lot, also with the null option I guess) and create a static const string for the "null" literal so it has to be created only once.

sebaszm: std::string and no T() was used on purpose as the underlying buffer for the json string (_value) is defined as a std::string, not string. The idea was to keep both the same. 

The reference to a temporary string is safe here, but as pointed out by Marcel, the repeated string construction might not be optimal.

sebaszm: The underlying buffer (_value) is a std::string as well, hence it's on purpose kept the same here. If this is incorrect it needs to be changed too.

Agree, I'll update this to construct the null string object once as some cycles could definitely be saved, but I don't think it would make a drastic difference as any decent STL implementation will use SOO for a 4-char string, thus omitting heap allocation altogether.

Alternatively a bare-metal memcpy would most likely outclass any string-based attempt easily... ;)

MFransen69: Agree that in more modern stl versions SOO will kick in, in this case (older GCC stl versions do not have this, do not know by heart what version introduced it). But then still the bytes will need to be copied into the SOO buffer for that string.

pwielders: You are right, this is the correct implementation. It is a design decission that the JSON is alwatys serialized (and deserialized) from multibyte characters (hence the std::string and the missing _T macros) So it is oke. I was just talking rubbish :-)

pwielders: See my previous comment on _T std::string, Do not change that Sebas might the right decision.

pwielders:
But the million dolar question is, did this small change fix the JSONRPC inconsistency in third party apps? So can we close the ticket with this pull ?

pwielders:
I will already merge this as functionality wise it is oke. The optimization can be done in antoher pull request, but I would like this to be resolved before any one else runs into it.

PR #102, opened 2019-04-17
[JSON] Optimized 565d505

sebaszm:
Optimize a bit.


pwielders:
No we did the havy lifting it is up to marcel to prove that this is faster :-)

sebaszm:
Hopefully the static initialisation guard code does not eat up what was gained. ;)

PR #105, opened 2019-04-25
Sebaszm/rpc json controller

sebaszm:
Add JSON-RPC endpoints and notifications to the Controller.
Note: the event "all" is still to be taken care of.

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -524,6 +524,7 @@ namespace PluginHost
     /* virtual */ void Server::Service::Notify(const string& message)
     {
         const string fullMessage("{\"callsign\":\"" + PluginHost::Service::Callsign() + "\", \"data\": " + message + " }");
+        printf("******** %s: %sn", PluginHost::Service::Callsign(), message);
 
         // Notify the base class and the subscribers
         PluginHost::Service::Notification(message);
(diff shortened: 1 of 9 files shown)


Source/WPEFramework/PluginServer.cpp
```diff
@@ -524,6 +524,7 @@ namespace PluginHost
     /* virtual */ void Server::Service::Notify(const string& message)
     {
         const string fullMessage("{\"callsign\":\"" + PluginHost::Service::Callsign() + "\", \"data\": " + message + " }");
+        printf("******** %s: %sn", PluginHost::Service::Callsign(), message);
```
pwielders: Dont't know if this one is gone in the next commit, obut if not can we take this line out

Source/WPEFramework/PluginServer.cpp
```diff
@@ -524,7 +524,6 @@ namespace PluginHost
     /* virtual */ void Server::Service::Notify(const string& message)
     {
         const string fullMessage("{\"callsign\":\"" + PluginHost::Service::Callsign() + "\", \"data\": " + message + " }");
-        printf("******** %s: %sn", PluginHost::Service::Callsign(), message);
```
pwielders: Wow, faster than light :-) Oke ignore my last remark :-)

pwielders:
Looks good. Will merge! just one change required but you can do that in a seperate change set. WOuld like to move the escaping of the string character to the base class of the JSON::String object. This is probably usefull in may more areas. Would like to avoid that everyone will do this whenever the JSON string needs to be escaped.

PR #109, opened 2019-04-29
[jsonrpc] Add jsonrpc to WifiControl

pawel-stanek-ml:


pawel-stanek-ml:
Investigating Resfdull API breakage

PR #112, opened 2019-04-30
add keyID in adapter_session_decrypt() call

goruklu:


goruklu:
Clients (wpe-webkit) using adapter_session_decrypt() calls need to be updated when the above commit is merged.

goruklu:
replaced with #117

PR #116, opened 2019-05-08
[JSON-RPC] Construct text message out of error code for failed RPC calls

sebaszm:
Some template witchery but allows to maintain a single error code list.


Source/core/Portability.h
```diff
@@ -604,48 +604,78 @@ namespace Core {
...
-    const uint32_t ERROR_FIRST_RESOURCE_NOT_FOUND = 34;
-    const uint32_t ERROR_SECOND_RESOURCE_NOT_FOUND = 35;
-    const uint32_t ERROR_ALREADY_RELEASED = 36;
-    const uint32_t ERROR_NEGATIVE_ACKNOWLEDGE = 37;
-    const uint32_t ERROR_INVALID_SIGNATURE = 38;
-    const uint32_t ERROR_READ_ERROR = 39;
-    const uint32_t ERROR_WRITE_ERROR = 40;
-    const uint32_t ERROR_INVALID_DESIGNATOR = 41;
+    template<uint32_t N>
+    inline const char *_Err2Str() {
```
pwielders: Can we make this const TCHAR* i.s.o. char.

Source/core/Portability.h
```diff
@@ -604,48 +604,78 @@ namespace Core {
...
+    ERROR_CODE(ERROR_FIRST_RESOURCE_NOT_FOUND, 34);
+    ERROR_CODE(ERROR_SECOND_RESOURCE_NOT_FOUND, 35);
+    ERROR_CODE(ERROR_ALREADY_RELEASED, 36);
+    ERROR_CODE(ERROR_NEGATIVE_ACKNOWLEDGE, 37);
+    ERROR_CODE(ERROR_INVALID_SIGNATURE, 38);
+    ERROR_CODE(ERROR_READ_ERROR, 39);
+    ERROR_CODE(ERROR_WRITE_ERROR, 40);
+    ERROR_CODE(ERROR_INVALID_DESIGNATOR, 41);
+
+    constexpr uint32_t ERROR_LAST = 41;
```
pwielders: Why is this needed, this will probably lead to issues as people forget to update the ERROR_LAST and I do not understand what it brings ?
Potentially you could in the default template above, convert the number to a string, and I gues the error to string works for all numbers that have not been explicitly casted to the text.

Source/core/Portability.h
```diff
@@ -604,48 +604,78 @@ namespace Core {
...
-    const uint32_t ERROR_SECOND_RESOURCE_NOT_FOUND = 35;
-    const uint32_t ERROR_ALREADY_RELEASED = 36;
-    const uint32_t ERROR_NEGATIVE_ACKNOWLEDGE = 37;
-    const uint32_t ERROR_INVALID_SIGNATURE = 38;
-    const uint32_t ERROR_READ_ERROR = 39;
-    const uint32_t ERROR_WRITE_ERROR = 40;
-    const uint32_t ERROR_INVALID_DESIGNATOR = 41;
+    template<uint32_t N>
+    inline const char *_Err2Str() {
+        return _T("");
```
pwielders: make the number convert to a static char array here, for all not properly explicitly spe'ed template instances ehich, agree brings a race condition but than maybe you should always return a std::string (typed string) ?

Source/core/Portability.h
pawel-stanek-ml: I think this is even required due to _T() usage.

Source/core/Portability.h
```diff
@@ -604,48 +604,78 @@ namespace Core {
...
+    {
+        return (code == N? _Err2Str<N>() : _bogus_ErrorToString<N-1>(code));
+    };
+
+    template<>
+    inline const char* _bogus_ErrorToString<0ul>(uint32_t code) 
+    {
+        return (code == 0? _Err2Str<0ul>() : _Err2Str<~0ul>());
+    };
+
```
pwielders: Might be interesting to see how much code is generated for this and if the compiler can really optimize this.....

The recursive strawl down to the proper template instantiation is nice, but if I understand correctly, error code 1 to text is a 40 deep callstack :-)

Source/core/Portability.h
pawel-stanek-ml: This  is a start condition for the template recursion (with +specialization for 0 being a stop condition). I was trying to figure out how to avoid it too.

Just for te sake of the discussion I present how I would do it:
Imagine file errors.inc with the following content:
ERROR(ERROR_NONE, 0)
ERROR(ERROR_INPROGRESS, 12)
ERROR(ERROR_WRITE_ERROR, 40)
ERROR(ERROR_INVALID_DESIGNATOR, 41)
... and so on

Now in Portability.h imagine:

#define ERROR(CODE, VALUE) const uint32_t CODE = VALUE;

#include "errors.inc"

#undef ERRORS

struct ErrorToCodeMapEntry {
 uint32_t code;
 char* desc;
};

#define ERROR(CODE, VALUE) \
{ ERROR, #ERROR},

ErrorToCodeMapEntry entries[] = {
#include "errors.inc"
};

#undef ERROR

+ the functions which finds the map entry givent the code and returns the string for it.

pwielders: If we are going for a map, I suggest making it an enum.

pwielders: because enum Core::Enumerate<> template already has a value to name tranlation using a table.

Source/core/Portability.h
sebaszm: See here: [https://godbolt.org/z/XYMUli](https://godbolt.org/z/XYMUli)
Ended up with a switch/case-like code.

Source/core/Portability.h
sebaszm: ERROR_LAST is a big drawback of this solution, and I don't see a way of getting rid of it as it needs to be a compile time constant. If we change the error codes to an enum, none of the template trickery would even be needed. And if we assume the numbers need to be consecutive, then it's even simpler, just using a lookup table.

pawel-stanek-ml: I like Enumerate suggestion too.

Source/core/Portability.h
pwielders: Well @sebaszm you keep on amazing me :-) First the interesting template construction, than the compiler output page.
For me this is good to go know, it is not a 40 callstack deep code !

pwielders: Hippieware rules, commercial stuff sucks..
Also looked what the VisualStudio Compiler does it the hard way. 5411 (lines of code) versus 307 (by GCC) 
So on windows it sucks but fortunatly that is not our target platform.

sebaszm: Wow, they are really pathetic then! Frankly I would have not expect a modern compiler to trip over this code. But the gcc output is a bit of a let down too, because I hoped it would notice a jump table opportunity as all the cases are contiguous. All in all, perhaps indeed this approach is a tad too esoteric, especially for embedded systems, where you never know how lousy compiler you'll have to use.

sebaszm: The winner of this contest is clang (190 lines). Did a lookup table. :)
https://godbolt.org/z/WZHihK

pwielders: Oke all in all, and knowing that the platform will never be compiled by a microsoft compiler for production platforms, I am fin with both options (enum with default Core::enumeratetype<> lookup functionality) or the presented solution above. Lets have your votes?

pawel-stanek-ml: I'm biased becaues I presented how I would approach that and the way I see it is closer to enumeratetype<> than this template meta programming. OTOH  - correct me if I'm wrong - using enumeratetype<> would require a lot of places to be changed so that the error Value() is called in places error code is required? If I'm wrong I'd say let's go with this. Otherwise if - using the enumeratetype<> is transparent and cheap I'd still consider it. The advantage of this solution is: it's ready with no to little work (time)  required.:).

pawel-stanek-ml: Another idea which I find inferior yet allowing to have full control independent on the compiler is to run python script (compile time) to generate the header with Code2Description() implementation :D. It'd use switch-case

MFransen69:
Sorry only catching up now.

If I'm right the whole bogus construction (no pun intended I didn't choose the name ;) ) could be replaced by a constexpr function in C++ 14. (not yet in C++11 as no loops are allowed). Would be interesting to see what that would change to the generated code (certainly in Visual Studio if I understood correctly). But it will be a few years before we move to C++14 probably ;)

Best, Marcel


On Thu, 2019-05-09 at 04:27 -0700, pawel-stanek-ml wrote:

@pawel-stanek-ml commented on this pull request.

In Source/core/Portability.h<https://github.com/WebPlatformForEmbedded/WPEFramework/pull/116#discussion_r282444982>:

> +    ERROR_CODE(ERROR_INVALID_DESIGNATOR, 41);

+

+    constexpr uint32_t ERROR_LAST = 41;

+

+    template<uint32_t N>

+    inline const char* _bogus_ErrorToString(uint32_t code)

+    {

+        return (code == N? _Err2Str<N>() : _bogus_ErrorToString<N-1>(code));

+    };

+

+    template<>

+    inline const char* _bogus_ErrorToString<0ul>(uint32_t code)

+    {

+        return (code == 0? _Err2Str<0ul>() : _Err2Str<~0ul>());

+    };

+


Another idea which I find inferior yet allowing to have full control independent on the compiler is to run python script (compile time) to generate the header with Code2Description() implementation :D. It'd use switch-case

Source/core/Portability.h
pwielders: I am not a real-fan of Yet Another Code Generator for this issue.
So options 1 (the template ) and option 2 the Core:EnumeratType<> are the only feasable options. Leave it up to @sebaszm to take a pick.

sebaszm: Well, the template is the code generator here. :) 
I have fixed the TCHAR and moved to enum with an x-macro idiom to get rid of biggest drawback, - - the ERROR_LAST variable. I think it's good enough now!

Source/core/Portability.h
```diff
@@ -604,48 +604,91 @@ namespace Core {
...
+        ERROR_CODE(ERROR_SURFACE_UNAVAILABLE, 32) \
+        ERROR_CODE(ERROR_PLAYER_UNAVAILABLE, 33) \
+        ERROR_CODE(ERROR_FIRST_RESOURCE_NOT_FOUND, 34) \
+        ERROR_CODE(ERROR_SECOND_RESOURCE_NOT_FOUND, 35) \
+        ERROR_CODE(ERROR_ALREADY_RELEASED, 36) \
+        ERROR_CODE(ERROR_NEGATIVE_ACKNOWLEDGE, 37) \
+        ERROR_CODE(ERROR_INVALID_SIGNATURE, 38) \
+        ERROR_CODE(ERROR_READ_ERROR, 39) \
+        ERROR_CODE(ERROR_WRITE_ERROR, 40) \
+        ERROR_CODE(ERROR_INVALID_DESIGNATOR, 41)
```
pawel-stanek-ml: Personally I'd prefer it in a separate file without need to use '\' at the line end and inclusion of that file here as in my example.

Source/core/Portability.h
pwielders: Do like the fact that the END_MARKER, is now automatically calculated, That I think was the biggest negative point from the template solution. 
My remark on the "Yet Another Code Generator" was on creating another PERL/Python (External) script to generate code.
Do Like templates. 
For me good to merge

PR #119, opened 2019-05-14
Constify OnKeyMessage() f_pszUrl parameter

dwrobel:
Helps to fix errors on OCDM-Playready like;
 error: ISO C++ forbids converting a string constant to ‘char*’
         m_piCallback->OnKeyMessage((const uint8_t *) "", 0, "");


dwrobel:
Requires also: https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/pull/116

PR #122, opened 2019-05-23
Generated OpenCDM proxy stubs

sebaszm:
Generated OCDM proxystubs. 

Please verify the interface whether the in/out tags are defined properly, in particular the GetSecureStop, GetSecureStopIds and StoreLicenseData methods.

Btw, do we still need the clunky ocdm_announce_proxy_stub thing, as we force --no-as-needed anyway?


pwielders:
Looks good. If it work, for me it is a go. The method ocdm_announce_proxy_stub is only there for windows :-) Windows needs to "see" that a method is actually being called in the dynamically loaded module and other wise it will optimize it out. The Server side (also buildable under windows) call this method so windows will not optimize it out. So as windows is not really supported it is oke to leave it out.

PR #124, opened 2019-05-28
[compositorclient] Make westeros implementation to use the proper di…

pawel-stanek-ml:
…play id.

Try to start straighting things with CLIENT_IDENTIFIER up as well (WPE-300).

### DIFF
--- a/Source/compositorclient/Client.h
+++ b/Source/compositorclient/Client.h
@@ -46,7 +46,47 @@ namespace Compositor {
             virtual int32_t Height() const = 0;
         };
 
+        // If CLIENT_IDENTIFIER variable is set its value is tokenized using comma as the separator
+        // and the first token is used to override the passed in name. The second - if present - is
+        // used as an underlaying compositor specific identifier.
+        // See GetOverrides()
         static IDisplay* Instance(const std::string& displayName);
+        static bool GetOverrides(std::string* name, std::string* identifier)
+        {
+            ASSERT(name != nullptr || identifier != nullptr);
+
+            const char* overrides = getenv("CLIENT_IDENTIFIER");
+            if (overrides != nullptr && overrides[0] != '\0') {
+                const char* delimiter = strchr(overrides, ',');
+                if (delimiter != nullptr) {
+                    ptrdiff_t delimiterPos = delimiter - overrides;
+                    if (name != nullptr) {
+                        name->clear();
+                        name->insert(0, overrides, delimiterPos);
+                    }
+
+                    if (identifier != nullptr) {
+                        identifier->clear();
+                        if (delimiterPos + 1 < strlen(overrides))
+                            identifier->insert(0, overrides + delimiterPos + 1);
+                    }
+                } else {
+                    // No delimiter found so it's only a name.
+                    if (name != nullptr) {
+                        name->clear();
+                        name->insert(0, overrides);
+                    }
+
+                    if (identifier != nullptr) {
+                        identifier->clear();
+                    }
+                }
+
+                return true;
+            }
+
+            return false;
+        }
 
         virtual ~IDisplay() {}
 
--- a/Source/compositorclient/Nexus/Implementation.cpp
+++ b/Source/compositorclient/Nexus/Implementation.cpp
@@ -71,13 +71,6 @@ namespace Nexus {
                 , _keyboard(nullptr)
             {
 
-                uint32_t nexusClientId(0); // For now we only accept 0. See Mail David Montgomery
-                const char* tmp(getenv("CLIENT_IDENTIFIER"));
-
-                if ((tmp != nullptr) && ((tmp = strchr(tmp, ',')) != nullptr)) {
-                    nexusClientId = atoi(&(tmp[1]));
-                }
-
                 NXPL_NativeWindowInfoEXT windowInfo;
                 NXPL_GetDefaultNativeWindowInfoEXT(&windowInfo);
                 windowInfo.x = 0;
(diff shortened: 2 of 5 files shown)


Source/compositorclient/Client.cpp
```diff
@@ -0,0 +1,39 @@
...
+                name->clear();
+                name->insert(0, overrides, delimiterPos);
+            }
+
+            if (identifier != nullptr) {
+                identifier->clear();
+                if (delimiterPos + 1 < strlen(overrides))
+                    identifier->insert(0, overrides + delimiterPos + 1);
+            }
+        }
```
pawel-stanek-ml: Lacks else

Source/compositorclient/Client.h
```diff
@@ -46,7 +46,12 @@ namespace Compositor {
             virtual int32_t Height() const = 0;
         };
 
+        // If CLIENT_IDENTIFIER variable is set its value is tokenized using comma as the separator
+        // and the first token is used to override the passed in name. The second - if present - is
+        // used as an underlaying compositor specific identifier.
+        // See GetOverrides()
         static IDisplay* Instance(const std::string& displayName);
+        static bool GetOverrides(std::string* name, std::string* identifier);
```
pwielders: Why do we need this on the public interface. I guess they can be kept in the implementation no one needs to externally work with them, I hope, and otherwise lets discuss on skype.

pwielders: Oke, clear. See the remainder of the code, this method just splits the name and the "remainder"

pwielders: Slight preference to keep the code in the header file. This way the interface stays header file only. Now the interface contains code...

Source/compositorclient/Nexus/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@ message("Setup ${TARGET} v${VERSION}...")
 find_package(EGL REQUIRED)
 find_package(NXCLIENT REQUIRED)
 
-add_library(${TARGET} SHARED Implementation.cpp)
+add_library(${TARGET} SHARED
+            Implementation.cpp
+            ../Client.cpp)
```
pwielders: And it avoids that specific implementations need to pick up interface header file code..
(See remark with the ICLient interface)

pawel-stanek-ml:
Note that this should allow to simplify the sites using compositorclientlik here: https://github.com/WebPlatformForEmbedded/WPEBackend-rdk/commit/6ada0dc4763d31a0804a9084e883bda7f2c52757
Similar could be done for Netflix etc but should work even if not done.

HaseenaSainul:
> Note that this should allow to simplify the sites using compositorclientlik here: [WebPlatformForEmbedded/WPEBackend-rdk@6ada0dc](https://github.com/WebPlatformForEmbedded/WPEBackend-rdk/commit/6ada0dc4763d31a0804a9084e883bda7f2c52757)
> Similar could be done for Netflix etc but should work even if not done.

@pawel-stanek-ml, yes you are correct, the Westeros will work without display name/id. If we are giving the name/id as NULL, it will try to get the currently running compositor. Other wise it will connect only if the compositor name is matching. In most of the Comcast sources, they are using the display name as null while connecting it to the compositor.
If we are moving the display name/id parsing logic to client, we have to update Netflix code as well (https://github.com/Metrological/netflix/blob/5.1.1-dev-latest/netflix/src/platform/gibbon/WPECompositor.h#L101)

pawel-stanek-ml:
> @pawel-stanek-ml, yes you are correct, the Westeros will work without display name/id. If we are giving the name/id as NULL, it will try to get the currently running compositor. Other wise it will connect only if the compositor name is matching. In most of the Comcast sources, they are using the display name as null while connecting it to the compositor.
> If we are moving the display name/id parsing logic to client, we have to update Netflix code as well (https://github.com/Metrological/netflix/blob/5.1.1-dev-latest/netflix/src/platform/gibbon/WPECompositor.h#L101)

@HaseenaSainul Yes. Comcast uses nullptr all the time because wl_display_connect() - if no name is given - looks at WAYLAND_DISPLAY env variable (which is used in RDK to specify display other then wayland-0, or tries wayland-0 by default.
Also note that Netflix or Webkit (I haven't seen Cobalt or Amazon but I'd imagine they do the same since they use compositor client as well) didn't really have a parsing logic which was moved. They parsed the name out of the var to pass it to the compositor and the compositor (nexus only til this moment) was parsing the id out of it. I made the compositor to parse the var and use it as overrides to what was passed via the API. This means the code in Webkit, Netflix etc can be cleaned up but it will not break if not done immediately.

HaseenaSainul:
> > @pawel-stanek-ml, yes you are correct, the Westeros will work without display name/id. If we are giving the name/id as NULL, it will try to get the currently running compositor. Other wise it will connect only if the compositor name is matching. In most of the Comcast sources, they are using the display name as null while connecting it to the compositor.
> > If we are moving the display name/id parsing logic to client, we have to update Netflix code as well (https://github.com/Metrological/netflix/blob/5.1.1-dev-latest/netflix/src/platform/gibbon/WPECompositor.h#L101)
> 
> @HaseenaSainul Yes. Comcast uses nullptr all the time because wl_display_connect() - if no name is given - looks at WAYLAND_DISPLAY env variable (which is used in RDK to specify display other then wayland-0, or tries wayland-0 by default.
> Also note that Netflix or Webkit (I haven't seen Cobalt or Amazon but I'd imagine they do the same since they use compositor client as well) didn't really have a parsing logic which was moved. They parsed the name out of the var to pass it to the compositor and the compositor (nexus only til this moment) was parsing the id out of it. I made the compositor to parse the var and use it as overrides to what was passed via the API. This means the code in Webkit, Netflix etc can be cleaned up but it will not break if not done immediately.

@pawel-stanek-ml, I am fine with your inputs :-)

PR #128, opened 2019-06-07
BRCM: Secbuf usage change to set input buffer in GLR, not in CRR

jinlee-brcm:
If the key control block does not require a secure output buffer, we restrict the input buffer to be only GLR which can be accessed from the HOST processor. So the chage is that Webkit provides encrypted data in GLR, and then CDM provides back decrypted data in CRR which can't be accessed from the HOST processor.

### DIFF
--- a/Source/ocdm/adapter/broadcom-svp/open_cdm_adapter.cpp
+++ b/Source/ocdm/adapter/broadcom-svp/open_cdm_adapter.cpp
@@ -7,6 +7,7 @@
 #include <b_secbuf.h>
 #include <gst_brcm_svp_meta.h>
 #include <assert.h>
+#include "Trace.h"
 
 struct RPCSecureBufferInformation {
     uint32_t type;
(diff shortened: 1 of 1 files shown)


pwielders:
Is there a document you can share on what this change means. The abbreviations GLR/CRR are a bit of a mystery to me :-)
Could you change the fprintf to TRACE_L1() macros. This allows us to control the output in the different builds. 
We have a code guideline that requests to have only 1 return statement per method. The rationale behind this is Maintainebility. If a next person needs to ehance/change/fix thjis code, he needs to validate all "return" paths. If there is only 1 it is easier to maintina/enhance.

jinlee-brcm:
Generally we don't have a document regarding security. Here is what the acronyms stand for.

CRR: Compressed Restricted Region, which Host CPU can not access, but Secure Processor can access
GLR: Global Region, which both Host CPU and Secure Processor can access

I will update the change according to your guideline.

pwielders:
Good to go!

Issue #130, opened 2019-06-14
Opencdm Update API Timeout/Lockup issues for widevine.


SR-Lavanyaa:
**branch:** master , WPEFramework 
 
**Reproducible with commit Id** - <62dfa673835c962dbf825eee11dfe9a3429c934f>
using https://github.com/WebPlatformForEmbedded/OCDM-Widevine plugin(ac3a765bdf96027fcb4ff67b0ca69858eb5dd665).
 
In Master Branch the Source File Location:  WPEFramework/Source/ocdm/ExtendedOpenCDMSession.h
In the tested commit SHA - Source File Location:   WPEFramework/Source/ocdm/open_cdm.cpp:
 
**Issue:**
 
In cases where the License Server or application provides an incorrect license response blob to the opencdm update API the caller thread gets blocked infinitely. OpenCdm::Update API goes into a wait state with infinite wait.
__state.WaitState(SESSION_UPDATE | SESSION_MESSAGE, WPEFramework::Core::infinite);_
 
The WaitState() function inside OpenCDM::Update API , Not able to acquire the second Lock() , it seems some other thread has already acquired the lock() and it hangs inside while loop.
 
 
**Note:** 1. Tested this scenario using RELEASE builds.
                2. In Update API the value of [WPEFramework::Core::infinite] is "-1".
                3. With a code walkthough of Wpeframework ,we could see several instances of Wait Timeouts being changed from Infinite to 20 sec (eg) ,
but unfortunately such a switch seems to be missing for the OpenCDM::Update API flows.
 
Since Opencdm::Update API is called from our main media player process thread any erroneous invocation of the API results in lockup of the mediaplayer. 

**CODE SNIPPET:**
 
**WPEFramework/Source/ocdm/open_cdm.cpp:**
 
KeyStatus Update(const uint8_t* pbResponse, const uint16_t cbResponse, std::string& response) {
 
        _state = static_cast<sessionState>(_state & (~(SESSION_UPDATE|SESSION_MESSAGE)));
 
        OpenCDMSession::Update(pbResponse, cbResponse);
 
        _state.WaitState(SESSION_UPDATE | SESSION_MESSAGE, WPEFramework::Core::infinite);
        if ((_state & SESSION_MESSAGE) == SESSION_MESSAGE) {
                response = "message:" + _message;
        }
 
        return CDMState(_key);
}

pwielders:
[...]

PR #134, opened 2019-06-19
Opencdm Update API Timeout/Lockup issues.

SR-Lavanyaa:
For Fixing Issue #130
OpenCdm::Update API goes into a wait state with infinite wait.
In cases where the License Server or application provides
an Incorrect License Response blob to the opencdm update API
the caller thread gets blocked infinitely.

### DIFF
--- a/Source/ocdm/ExtendedOpenCDMSession.h
+++ b/Source/ocdm/ExtendedOpenCDMSession.h
@@ -204,9 +204,15 @@ public:
         _state = static_cast<sessionState>(_state & (~(SESSION_UPDATE | SESSION_MESSAGE)));
 
         OpenCDMSession::Update(pbResponse, cbResponse);
+	// If we build in release, we do not want to "hang" forever, forcefull close after 20S waiting...
+	#ifdef __DEBUG__
+	unsigned int a_Time = WPEFramework::Core::infinite;
+	#else
+	unsigned int a_Time = 20000;// Expect time in MS
+	#endif
 
         _state.WaitState(SESSION_UPDATE | SESSION_MESSAGE,
-            WPEFramework::Core::infinite);
+            a_Time);
         if ((_state & SESSION_MESSAGE) == SESSION_MESSAGE) {
             response = "message:" + _message;
         }
--- a/Source/ocdm/ExtendedOpenCDMSession.h
+++ b/Source/ocdm/ExtendedOpenCDMSession.h
@@ -156,9 +156,15 @@ public:
     {
 
         ASSERT(IsValid() == true);
+	// If we build in release, we do not want to "hang" forever, forcefull close after 5S waiting...
+        #ifdef __DEBUG__
+        unsigned int a_Time = WPEFramework::Core::infinite;
+        #else
+        unsigned int a_Time = 5000;// Expect time in MS
+        #endif	
 
         _state.WaitState(SESSION_MESSAGE | SESSION_READY,
-            WPEFramework::Core::infinite);
+            a_Time);
 
         if ((_state & SESSION_MESSAGE) == SESSION_MESSAGE) {
             challenge = _message;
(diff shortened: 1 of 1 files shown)


pwielders:
The analyses is correct and the picture is much bigger than this. Prefer to do this face-to-face as this require some interaction and otherwise this story will be come a book :-)
To start with, not all design decisions in this code are Metrologicals. It is taken from Frauenhofer as a refernece implementation and was already integrated in WebKit, so we are facing an already deployed solution and I (as Metrological) can not explain why Fraunehofer opted for some of these design decisions. This blocking call, is one of them, however, analyzing the full picture, it does not harm the current approach. 
On the other hand the proposed change also does not harm the full picture as 20S in a computer time is pretty much infinite (same approach as we take on COMRPC, Release 3 second timeout, debug infinite).
But I have two issues with the proposed solution:
1) If you ran into this issue it is actually hiding a more fundamental issue. Within WebKit (consumer of OCDM) we have never run into this, because if playback  does not start, I think, within 6 secondens after the update of the License acquired, WebKit breaks down the pipeline and thus the OCDM session and within 6 Seconds + some processing time the whole is reset. If you encounter this issue somewhere in the code, you should cater for a scenario that playback could potentially fail (you get no fragment because server is down ??)   and should tear down the player (and thus the OCDM session).
2) I do understand that testing does not always "catch" all issues so why not, if possible, have some kind of safety net to always, at least process, even if it would be after 20S. Always better than a hanging system and maybe the rootcause of issue 1, although it hides a "real" issue occurs but was never found, should not stop the system from never creating a new OCDM session, if that would happen (doubt it because I think it becomes a leak, only after 16 leaks, your system can no longer play anymore). Anyhow, it still becomes a novel... but bottom-line I can see the reasoning for this patch. To accept the patch, I would like to have a symmetrical interface.  The Update method is not the only method that applies these infinite wait semantics, also have a look at (Load/Remove and maybe some more) they should be adapted as well to show the same behavior.

So what I am trying to say :-) Make sure there is no hidden real issue, if you are sure there is no real issue, the patch is not nessecary, if you can not oversee the full picture a fallback like this might be oke (willing to take the patch) but than make sure that all methods behave like that (so adapt at least the Load and Remove methods as well)

pwielders:
@SR-Lavanyaa somewhere I saw a response related to this pull request, unfortunately I can not find it anymore. Could you point me to that response ?

SR-Lavanyaa:
@pwielders Please find our response related to this pull request in Issue #130

SR-Lavanyaa:
@pwielders kindly review the updated commit.

woutermeek:
@SR-Lavanyaa please only define the variable a_Time only once, and re-use the same variable.

Source/ocdm/ExtendedOpenCDMSession.h
```diff
@@ -184,8 +190,14 @@ struct ExtendedOpenCDMSession : public OpenCDMSession {
         response.clear();
 
         if (OpenCDMSession::Load() == 0) {
+  	     // If we build in release, we do not want to "hang" forever, forcefull close after 5S waiting...
+             #ifdef __DEBUG__
```
woutermeek: Please only define this once

Source/ocdm/ExtendedOpenCDMSession.h
```diff
@@ -204,9 +216,15 @@ struct ExtendedOpenCDMSession : public OpenCDMSession {
         _state = static_cast<sessionState>(_state & (~(SESSION_UPDATE | SESSION_MESSAGE)));
 
         OpenCDMSession::Update(pbResponse, cbResponse);
-
-        _state.WaitState(SESSION_UPDATE | SESSION_MESSAGE,
-            WPEFramework::Core::infinite);
+	// If we build in release, we do not want to "hang" forever, forcefull close after 5S waiting...
+	#ifdef __DEBUG__
+	unsigned int a_Time = WPEFramework::Core::infinite;
```
woutermeek: Please only define this once

Source/ocdm/ExtendedOpenCDMSession.h
```diff
@@ -220,8 +238,14 @@ struct ExtendedOpenCDMSession : public OpenCDMSession {
         _state = static_cast<sessionState>(_state & (~(SESSION_UPDATE | SESSION_MESSAGE)));
 
         if (OpenCDMSession::Remove() == 0) {
-
-            _state.WaitState(SESSION_UPDATE, WPEFramework::Core::infinite);
+	    // If we build in release, we do not want to "hang" forever, forcefull close after 5S waiting...
+             #ifdef __DEBUG__
```
woutermeek: Please only define this once

Source/ocdm/ExtendedOpenCDMSession.h
SR-Lavanyaa: updated and posted again.

PR #144, opened 2019-07-05
[CMAKE] Use PYTHON_EXECUTABLE provided by build system

wouterlucas:
This change is required to make use of OE's/Yocto python-native package (over the host python) in the nativesdk sysroot and include the jsonref package.

Signed-off-by: wouterlucas <wouter@wouterlucas.com>


pwielders:
@bramoosterhuis I assume this is oke? Agree?

PR #146, opened 2019-07-08
[virtualinput] Remove AdministerAndSendKey and use SendKey

bramoosterhuis:


pwielders:
@bramoosterhuis This is what we did a couple a weeks ago? Right?

PR #149, opened 2019-07-17
[Core]Fix 'min' memory measurement value in the response for not acti…

Damian-Danylko:
…ve plugins


Damian-Danylko:
Not merged. Fix it in a different way.

PR #154, opened 2019-07-23
[Compositor] Added JSON RPC to compositor

mipo57:


Source/interfaces/json/Compositor.json
```diff
@@ -0,0 +1,243 @@
...
+          "screenresolution_720p",
+          "screenresolution_720p50hz",
+          "screenresolution_1080p24hz",
+          "screenresolution_1080i50hz",
+          "screenresolution_1080p50hz ",
+          "screenresolution_1080p60hz",
+          "screenresolution_2160p50hz",
+          "screenresolution_2160p60hz"
+        ],
+        "example": "screenresolution_480i"
```
sebaszm: don't need to prefix the enum

Source/interfaces/json/Compositor.json
```diff
@@ -0,0 +1,243 @@
...
+          "description": "Callsign not found",
+          "$ref": "#/common/errors/firstresourcenotfound"
+        }, 
+        {
+          "description": "Failed to set/get geometry",
+          "$ref": "#/common/errors/general"
+        }
+      ]
+    },
+    "visible": {
```
sebaszm: Use hidden/visible enum instead of boolean

PR #159, opened 2019-07-24
[OCDM] allow for decryption without key ID

svdmaar:
@pwielders @modeveci , currently Netflix doesn't pass keyID to opencdm_gstreamer_session_decrypt (passes nullptr instead). Until the introduction of the GstBuffers this always worked. This proposed pull would restore the functionality. But... can this become an issue when switching to SVP? Should Netflix store the key ID for each session and pass along when calling opencdm_gstreamer_session_decrypt ?


pwielders:
@svdmaar that should not be an issue. The key if is just another way to identify the session. For PR the relation is always 1-1 so 1 session has 1 keyid. If the keyid is required, the only keyid associated with the PR session is used. This should be enforced by passing in a nullptr and length of a id == 0

svdmaar:
@pwielders , thanks for the clarification! Branch is merged

PR #162, opened 2019-07-25
[DropbearServerPlugin] Initial version for review

s-rapolu:
DropbearServer plugin implementation related changes in WPEFramework (Thunder) repository


Source/interfaces/IDropbearServer.h
```diff
@@ -10,8 +10,12 @@ namespace Exchange {
     struct IDropbearServer : virtual public Core::IUnknown {
```
pwielders: Why would you expose an implementation detail? The whole idea behind interface based developemnet is that you expose "what" you do not "how". I think the "what" here is an "SSHServer" and the "how" you realize it is a "DropBear" implementation. 
However, if you would use the  https://tinyssh.org/ do you think this interface would not work :-)
So I would expect an interface name lik ISecureShellServer.h

Source/interfaces/IDropbearServer.h
```diff
@@ -10,8 +10,12 @@ namespace Exchange {
...
         enum { ID = ID_DROPBEARSERVER };
 
-        virtual uint32_t StartService(const string& port) = 0;
+        virtual uint32_t StartService(const std::string& host_keys, const std::string& port_flag, const std::string& port) = 0;
```
pwielders: As you can already see by the updates, the start service is receiving aonfiguration information Why not adapt the schema applied on more interface, just pass the configuration as a JSON string (so a string). 
Than use Initialize (const string& configuration) 
to initialize and start your server correctly.
This would be in line with a lot of more services ran in Thunder.

Source/interfaces/IDropbearServer.h
```diff
@@ -10,8 +10,12 @@ namespace Exchange {
...
         virtual uint32_t StopService() = 0;
```
pwielders: As a symetrical counterpart, this would become Deinitialize()

Source/interfaces/IDropbearServer.h
```diff
@@ -10,8 +10,12 @@ namespace Exchange {
...
+        virtual uint32_t GetTotalSessions() = 0;
```
pwielders: I guess this method could be const and what is the difference between a:
uint32_t GetTotalSessions() const
uint32_t GetSessionCount() const

Source/interfaces/IDropbearServer.h
```diff
@@ -10,8 +10,12 @@ namespace Exchange {
...
+        virtual uint32_t GetSessionsCount() = 0;
+        //virtual uint32_t GetSessionsInfo(Core::JSON::ArrayType<JsonData::DropbearServer::SessioninfoResultData>& response) = 0;
+        virtual uint32_t CloseClientSession(uint32_t client_pid) = 0;
```
pwielders: Seeing a close with a Pid and a count. I suggest offering a IClients interface within this interface wit the following semantics:
struct IClients  :  virtual public Core::IUnknown {
    virtual void Reset() = 0;
    virtual bool IsValid() const  = 0;
    virtual bool Next() = 0;
    virtual string RemoteId() const = 0;
    virtual void Close() = 0;
    virtual void Count() const = 0;
}

This way you can expose the current amount of client sessions (indicated by Count()), and you can close any sessions with a remote host (identificed by RemoteId()) by calling Close() withoud exposing any implementation details.

Source/interfaces/IDropbearServer.h
s-rapolu: Agreed. I'll make the corresponding changes.

Source/interfaces/IDropbearServer.h
s-rapolu: Sure. I'll create an interface with ISecureShellServer.h and make changes to not expose the implementation detail.

Source/interfaces/IDropbearServer.h
s-rapolu: **GetTotalSession()** method shall provide the count of "total" sessions (either active or inactive) handled by the Dropbear Server since from it's uptime (or plugin activation).

**GetSessionCount()** method shall provide the count of only "active" sessions currently handled by Dropbear Server since from it's uptime (or plugin activation).

Sure, I'll make both these methods as const.

Source/interfaces/IDropbearServer.h
s-rapolu: Agreed. I'll implement these changes.

Source/interfaces/IDropbearServer.h
s-rapolu: Agreed.

Source/interfaces/IDropbearServer.h
pwielders: I do not think that GetTotalSession() has any added value on the interface. Suggest to drop it.
The GetSessionCount will be part of the Itartor for IIterator* ActiveSessions(); and as such can be extracted from there. Furthermore, I do want to hide the fact that you need a pid to "close" a client, which can be kept internally if you add the iterator.
Just learned that some operators are already looking at different Secure Shell functionalities. We really need to make this interface as generic as possible without losing the functionality needed) .

s-rapolu:
Changed DropbearServer interface to SecureShellServer and applied some of the feedback points.

PR #165, opened 2019-07-29
[POWER] Add JSON RPC interface

Damian-Danylko:


Source/interfaces/json/Power.json
```diff
@@ -0,0 +1,94 @@
...
+            }
+      }   
+    }   
+  },
+  "properties": {
+    "state": {
+      "summary": "Power state",
+      "readonly": true,
+      "params": {
+        "type": "object",
```
sebaszm: don't need to enclose the property within an object

Damian-Danylko: Done.

PR #168, opened 2019-08-06
[Core] Unregistering the handler in IPC server.

aswaunni:
The handlers are registered after client connection. But not unregistered after connection closes or in destructor and its causing ASSERT on destruction.
https://github.com/WebPlatformForEmbedded/Thunder/blob/7214bcfcf984de5703f54ec9e129be1a65e16a8d/Source/core/IPCConnector.h#L742

I have put a fix for that.

### DIFF
--- a/Source/core/IPCChannel.h
+++ b/Source/core/IPCChannel.h
@@ -230,6 +230,7 @@ namespace Core {
             Close(infinite);
 
             if (_clients.size() > 0) {
+                InternalCleanup();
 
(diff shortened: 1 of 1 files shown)


aswaunni:
I tried that but changed that because it wont work if we call the Unregister before (since it removes the handlers).

aswaunni:
But Unregister() shouldn't be called because it erases the handler from the list
https://github.com/WebPlatformForEmbedded/Thunder/blob/1e4edeccda6e0aa7dde4d22dbb3153ec9a6af402/Source/core/IPCChannel.h#L313

PR #177, opened 2019-08-19
[Netflix] Add JSON RPC API

Damian-Danylko:


Source/interfaces/json/Netflix.json
```diff
@@ -0,0 +1,96 @@
...
+        },
+        {
+          "description": "Command is not set",
+          "$ref": "#/common/errors/unknownkey"
+        }        
+      ]
+    }
+  },
+  "events": {
+    "visibilitychange": {
```
MFransen69: Is there a reason the playing event was not added?
And I guess the Resumed state update gets in via de "include StateControl.json"?

Damian-Danylko: Indeed resumed/suspended events are handle by this part of API
"include": {
    "statecontrol": {
      "$ref": "StateControl.json#"
    }
Playing event is added, I missed this one.

Source/interfaces/json/Netflix.json
```diff
@@ -0,0 +1,96 @@
...
+        "example": "DADDAUFV2=00000000000000000000000000000001"
+      },
+      "errors": [
+        {
+          "description": "Netflix is unavailable",
+          "$ref": "#/common/errors/unavailable"
+        }
+      ]      
+    },
+    "visibility": {
```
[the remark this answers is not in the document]
Damian-Danylko: Indeed resumed/suspended events are handled by this part of API
"include": {
"statecontrol": {
"$ref": "StateControl.json#"
}
In context of visibility Netflix interface is giving you only this API
virtual void SetVisible(bool visibility) = 0;
to set visibility there is no option to check this value outside implementation.
So for this reason I did only writeonly property. Is it clear now?

PR #186, opened 2019-08-22
[Json] Added escaping of special characters in json string,

mipo57:
Escaping is done according to RFC 7159. Now WebKit should have no problems with deserializing json strings 
This change is followup of Jira ticket WPE-320 (Unescaped json)

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -1187,6 +1187,48 @@ namespace Core {
                 }
             }
 
+        private:
+            /*
+                Serialization methods
+            */
+            inline void AddEscapedCharacter(char* outputStream, char escapeSymbol, uint16_t& index, const uint16_t maxLength) const
+            {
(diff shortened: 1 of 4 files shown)


pawel-stanek-ml:
Once \uXXXX support is in place please ale enable the unit test checking invalid \u sequence (it's already there but disabled)

pwielders:
I am afraid that this is a breaking change. The JSON deserialisation can be done form a single character input device (like serial ports). There is a testcase that validates that this still works (to be found here: https://github.com/WebPlatformForEmbedded/WPEFrameworkTests/blob/master/Functional/JSONTests/main.cpp#L162
Here deserialization is validated with respect to singel incoming characters. Can you validate that this still works ?

Further more, try to avoid scanf/snprintf functions. Could you see if you can parse this "short" synatx yourself?

PR #194, opened 2019-08-29
[Serialization] Function for deserializing from hex string

mipo57:


Source/core/Serialization.cpp
```diff
@@ -156,6 +156,20 @@ namespace Core {
         }
     }
 
+    void EXTERNAL FromHexString(const string& hexString, uint8_t* object, uint16_t& length, const uint16_t maxLength) {
+        ASSERT(object != nullptr || maxLength == 0);
```
pwielders: I think this conversion has a bit of unexpected behaviour. Or ASSERT (string.length() % 2 == 0); or properly handle it. I prefer the latter since I think that "BA4" => { 0x0B, 0xA4 }
You could also combine the length and maxlength, length (on incoming) is he maxLength, you write in it what is added. but agree that is a matter of taste, nut I always prefer to keep the paramter list as short as possible

mipo57: I agree with the first one. Still, previously it would be also a kind of an unexpected behavior I think - although probably a more sane one. Now it will be documented by tests :)

As for the number of arguments - I think that returning length (rather than void) will satisfy both of our tastes :)

pwielders: :+1:

pwielders:
Good to merge. Looks really good ! Just need to discuss with QA how to merge this. If QA will or we can.

PR #197, opened 2019-09-02
Fixes in core Cyclicbuffer

aswaunni:
1.Fix to differentiate between full and empty buffer.
    head == tail -> empty
    (head + 1) == tail -> full
2. Checking the OVERWRITE flag while writing to buffer.


pwielders:
It looks like a sensible change but since it is core, let test this thoroughly. Are there unit tests for this? Did they run?

PR #198, opened 2019-09-04
[JSON] Added Hex and Oct Test Cases to JSONParser test and fixed issu…

aswaunni:
1. Added Hexadecimal and Octal parsing tests to JSONParser Test.
2. Fixed issued in Hexadecimal and Octal parsing in core JSON.h.

### DIFF
--- a/Tests/core/test_jsonparser.cpp
+++ b/Tests/core/test_jsonparser.cpp
@@ -492,6 +492,54 @@ namespace Tests {
         });
     }
 
+    TEST(JSONParser, PositiveHexNumber)
+    {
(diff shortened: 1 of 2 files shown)


Tests/core/test_jsonparser.cpp
```diff
@@ -492,6 +492,54 @@ namespace Tests {
         });
     }
 
+    TEST(JSONParser, PositiveHexNumber)
```
pawel-stanek-ml: I wrapped UTs like this in QUIRKS_MODE define (which is defined)

PR #200, opened 2019-09-06
[NumberType] Fixed parsing issue in core Number.h.

aswaunni:
Fixed conversion issue by adding extra condition for Hexadecimal (hexadecimal base type number in 0x format was not getting parsed)

### DIFF
--- a/Source/core/Number.h
+++ b/Source/core/Number.h
@@ -452,7 +452,7 @@ namespace Core {
                 if ((Value == 0) && (*Text == '0') && (Base == BASE_UNKNOWN)) {
                     // Base change, move over to an OCTAL conversion
                     Base = BASE_OCTAL;
-                } else if ((Value == 0) && (toupper(*Text) == 'X') && (Base == BASE_OCTAL)) {
+                } else if ((Value == 0) && (toupper(*Text) == 'X') && ((Base == BASE_OCTAL) || (Base == BASE_HEXADECIMAL))) {
                     // Base change, move over to an HEXADECIMAL conversion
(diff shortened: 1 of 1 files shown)


Source/core/Number.h
```diff
@@ -452,7 +452,7 @@ namespace Core {
                 if ((Value == 0) && (*Text == '0') && (Base == BASE_UNKNOWN)) {
                     // Base change, move over to an OCTAL conversion
                     Base = BASE_OCTAL;
-                } else if ((Value == 0) && (toupper(*Text) == 'X') && (Base == BASE_OCTAL)) {
+                } else if ((Value == 0) && (toupper(*Text) == 'X') && ((Base == BASE_OCTAL) || (Base == BASE_HEXADECIMAL))) {
```
pwielders: I think this should be without the base == BASE_OCTAL. it should only check for (Base == BASE_HEXADECIMAL)

Source/core/Number.h
```diff
@@ -559,7 +559,7 @@ namespace Core {
...
```
pwielders: See previous comment

Source/core/Number.h
```diff
@@ -632,7 +632,7 @@ namespace Core {
...
```
pwielders: See previous comment.

Source/core/Number.h
aswaunni: When the number is of the form 0x, first we are checking for '0' and Base is set to BASE_OCTAL and hence that check.

pwielders: You are right. Again a good fix.

PR #201, opened 2019-09-06
WPE-243

HaseenaSainul:
Configure environment variables which are given in the config.json from WEFramework Process


Source/core/SystemInfo.cpp
```diff
@@ -519,16 +519,18 @@ namespace Core {
...
-    /* static */ bool SystemInfo::SetEnvironments(const Core::JSON::ArrayType<SystemInfo::Environment>& environments)
+    /* static */ void SystemInfo::SetEnvironments(const Core::JSON::ArrayType<SystemInfo::Environment>& environments)
     {
         bool status = true;
         Core::JSON::ArrayType<Environment>::ConstIterator index(environments.Elements());
         while ((index.Next() == true) && (status == true)) {
-            if ((index.Current().Env.IsSet() == true) && (index.Current().Value.IsSet() == true)) {
-                status = Core::SystemInfo::SetEnvironment(index.Current().Env.Value(), index.Current().Value.Value(), true);
+            if ((index.Current().Key.IsSet() == true) && (index.Current().Value.IsSet() == true)) {
+                status = Core::SystemInfo::SetEnvironment(index.Current().Key.Value(), index.Current().Value.Value(), true);
```
pwielders: You really want to stop if one failse? I think you should just issue a SYSLOG(Logging::Startup, _T("..... warning for that key and leve it up to the user to determine if it is serious but still process the others.

HaseenaSainul: This is outdated @Pierre, This was already corrected in the following commits

pwielders:
Looks good but I would like to use the substitute functionality also in from a pluginperspective. This means that we extend the Shell a method called string ConfigSubstitition(const string& input) = 0;
The implementation of this method would be
return (_environment.SubstituteVariables(_config, input));
This requires to make the class EnvironmentConfig to be an aggregate of the PluginServer and not a static class.
If you have any questions left, just let me know.

Source/WPEFramework/EnvironmentConfig.h
```diff
@@ -0,0 +1,94 @@
...
+            _variables.insert(std::make_pair("binding", [](const Server::Config& config) { return config.Binding.Value();}));
+            _variables.insert(std::make_pair("datapath", [](const Server::Config& config) { return config.DataPath.Value();}));
+            _variables.insert(std::make_pair("persistentpath", [](const Server::Config& config) { return config.PersistentPath.Value();}));
+            _variables.insert(std::make_pair("systempath", [](const Server::Config& config) { return config.SystemPath.Value();}));
+            _variables.insert(std::make_pair("volatilepath", [](const Server::Config& config) { return config.VolatilePath.Value();}));
+            _variables.insert(std::make_pair("proxystubpath", [](const Server::Config& config) { return config.ProxyStubPath.Value();}));
+        }
+
+    private:
+        static bool SubstituteVariables(const Server::Config& config, string& value)
```
pwielders: Might be interesting to make this public and offer this for any string. See general remark on the submit to put this on the IShell interface.

Source/WPEFramework/EnvironmentConfig.h
```diff
@@ -0,0 +1,94 @@
...
+                        status = Core::SystemInfo::SetEnvironment(index.Current().Key.Value(), value, index.Current().Override.Value());
+                        if (status != true) {
+                            TRACE_L1("Failure in setting Key:Value:[%s]:[%s]\n", index.Current().Key.Value().c_str(), index.Current().Value.Value().c_str());
+                        }
+                    }
+                }
+            }
+        }
+
+        static void Initialize(const Server::Config& config)
```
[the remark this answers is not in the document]
HaseenaSainul: Thanks @pierre for the comments, but here need config value to construct the lambda map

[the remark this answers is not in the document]
pwielders: Just try compiling it without, you will see it will compile as well :-)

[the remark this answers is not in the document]
pwielders: The lambda only requires it on executions so in the Substitute, but there you pass it.

Source/plugins/Service.h
```diff
@@ -472,6 +477,7 @@ namespace PluginHost {
         state _state;
         Config _config;
         string _errorMessage;
+        Environment _environment;
```
pwielders: Could you please make the Environment object a part of the Server. Now there is a copy of this on every plugin, so it gets quite bloated. The service can reach it parent (the server) and thus the single copy.

PR #202, opened 2019-09-06
[Timer] Added check against current time before scheduling.

aswaunni:
Added check against current time before scheduling because scheduling of past time was also calling Timed() function.


aswaunni:
If the expectation is a wakeup call, then this change won't be needed. Thank you.

PR #209, opened 2019-09-17
[compositorclient:RPI] avoid local variable in the inner block for la…

HaseenaSainul:
…mbda functions

### DIFF
--- a/Source/compositorclient/RPI/Implementation.cpp
+++ b/Source/compositorclient/RPI/Implementation.cpp
@@ -430,6 +430,7 @@ private:
     uint16_t _touch_x;
     uint16_t _touch_y;
     uint16_t _touch_state;
+    IDisplay::IKeyboard::state _key_state;
 
     mutable uint32_t _refCount;
 };
@@ -561,6 +562,7 @@ Display::Display(const string& name)
     , _touch_x(-1)
     , _touch_y(-1)
     , _touch_state(0)
+    , _key_state(IDisplay::IKeyboard::released)
     , _refCount(0)
 {
 }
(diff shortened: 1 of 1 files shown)


Source/compositorclient/RPI/Implementation.cpp
```diff
@@ -579,8 +581,8 @@ int Display::Process(const uint32_t data)
         std::function<void(SurfaceImplementation*)> action = nullptr;
 
         if (message.type == KEYBOARD) {
-            const IDisplay::IKeyboard::state state = ((message.keyData.type == KEY_RELEASED)? IDisplay::IKeyboard::released : IDisplay::IKeyboard::pressed);
```
MFransen69: Hi Haseena,

Could you perhaps explain why you are making this change? The state is only used locally in this method, correct?

HaseenaSainul: const IDisplay::IKeyboard::state state is local to this block and it is not accessible outside the block. i.e, here the lambda function action will get invoked outer this block (https://github.com/WebPlatformForEmbedded/Thunder/blob/db2e4b21499b837810d9780d96de045c442da11a/Source/compositorclient/RPI/Implementation.cpp#L620)  and that time this value will be junk. Right now I have kept _key_state as member variable, I can make it as the local to the https://github.com/WebPlatformForEmbedded/Thunder/blob/db2e4b21499b837810d9780d96de045c442da11a/Source/compositorclient/RPI/Implementation.cpp#L574 as well

MFransen69: Yes, sorry missed that it was used in a lambda so not only localy, didn't look any further then my nose is long ;) (Dutch saying for that I did not look long enough :) )
But then whey pass it by reference to the lambda and not value? (in the latter case it still can be kept local) Should the lambda use the value at creation of the lambda or is there a chance it is updateded later and the current value at execution of the lambda must be used?

HaseenaSainul: Here the lambda using only the value at the creation. But if we follow the latter case (pass by value), do we have to use a different function prototype for lambda, right ? In that case, we have touch all other lambda functions created for mouse / touch etc.

PR #211, opened 2019-09-17
Added support to return key id in key_update_callback.

neeraj-deshpande:


pawel-stanek-ml:
Why is this needed? The key ids being properly passed in is already on both master )done within multikey sessions) and on development/new-ocdm-interface (which will become a new master very soon - like this week).

neeraj-deshpande:
Recently I integrated OCDM (branch sander/new-ocdm-interface) with Cobalt browser where I am getting NULL key id in the key_update_callback. Here is the place where nullptr is hard-coded: 
https://github.com/WebPlatformForEmbedded/Thunder/blob/1b20353cf71d6807c2d40ea2ba109df5d305dcfd/Source/ocdm/open_cdm_impl.h#L992

pawel-stanek-ml:
My point is: rebase to the newest master or development/new-ocdm-interface which is soon to be master. It's already done there

pwielders:
@neeraj-deshpande I think this is resolved in the upcoming OCDM changes. If so can you close this pull-request.

pawel-stanek-ml:
This was resolved even without upcoming changes. This already works on master after WV multisessions

Issue #212, opened 2019-09-18
Request to fix compilation error in warrior branch, Faced it while building raspberrypi image using poky (warrior branch).


thadivaka-gopikrishna:
| FAILED: Source/compositorclient/Wayland/CMakeFiles/compositorclient.dir/Westeros.cpp.o
[...]

PR #213, opened 2019-09-18
[Compositorclient: Wayland]: nullptr changed to NULL to handle castin…

HaseenaSainul:
…g to EGLNativeWindowType (pointer and nonpointer types)

The actual type of EGLNativeWindowType is pointer or non pointer based on the platform used. So if we use the nullptr to do the casting, it always expect the casting to a pointer type only. As per the issue: https://github.com/WebPlatformForEmbedded/Thunder/issues/212, here the EGLNativeWindowType is 'long unsigned int', but in api build root case, it is void*. So I think better to use NULL, instead of nullptr.

### DIFF
--- a/Source/compositorclient/Wayland/Westeros.cpp
+++ b/Source/compositorclient/Wayland/Westeros.cpp
@@ -653,7 +653,7 @@ namespace Wayland {
                     _eglSurfaceWindow = eglCreateWindowSurface(
                         _display->_eglDisplay,
                         _display->_eglConfig,
-                        static_cast<EGLNativeWindowType>(nullptr),
+                        reinterpret_cast<EGLNativeWindowType>(NULL),
                         nullptr);
                 }
 
--- a/Source/compositorclient/Wayland/Westeros.cpp
+++ b/Source/compositorclient/Wayland/Westeros.cpp
@@ -653,7 +653,8 @@ namespace Wayland {
                     _eglSurfaceWindow = eglCreateWindowSurface(
                         _display->_eglDisplay,
                         _display->_eglConfig,
-                        reinterpret_cast<EGLNativeWindowType>(NULL),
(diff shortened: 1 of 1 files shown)


Source/compositorclient/Wayland/Westeros.cpp
```diff
@@ -653,7 +653,7 @@ namespace Wayland {
                     _eglSurfaceWindow = eglCreateWindowSurface(
                         _display->_eglDisplay,
                         _display->_eglConfig,
-                        static_cast<EGLNativeWindowType>(nullptr),
+                        reinterpret_cast<EGLNativeWindowType>(NULL),
```
MFransen69: Hi Haseena, may I ask why this change was made? In general I would prefer nullptr over NULL (more typesafe) and the use of static_cast over reinterpret_cast. Does this change fix a problem?

HaseenaSainul: Yes, to address the https://github.com/WebPlatformForEmbedded/Thunder/issues/212. The type of EGLNativeWindowType is either pointer or standard data type. In RPI - BR build it is void*, but as per the issue:212: it is long unsigned int. So if we go with static_cast<EGLNativeWindowType>(nullptr) for that case it will make compile error. Since nullptr always expected the casting to pointer. So due to the behaviour of EGLNativeWindowType I think better to use NULL, since it can be casted to both cases.

MFransen69: Ah, that makes sense! "Nice" that they are so consistent in their types ;) Ok, but I fully understand. Perhaps a good idea to add a comment to the code for the reason for the NULL, so other people who look at it will also understand and not change it back ;)

pwielders:
Discussed with Marcel. He will pick it up. Needs further investigation.

PR #215, opened 2019-09-20
[Portability] Fixed the memrcpy issue.

anjalirajan:
Fixed the memrcpy issue  in portability .


Source/core/Portability.cpp
```diff
@@ -174,6 +174,7 @@ void* memrcpy(void* _Dst, const void* _Src, size_t _MaxCount)
         *destination-- = *source--;
         --_MaxCount;
     }
+    *destination-- = *source--;
```
pwielders: good fix, but the decrements on both sides are not nessecary, so just do *destination == *source

PR #216, opened 2019-09-23
[Interfaces] Extend Streamer interface with custom events and stream …

sebaszm:
…elements

### DIFF
--- a/Source/interfaces/IStream.h
+++ b/Source/interfaces/IStream.h
@@ -14,8 +14,7 @@ namespace Exchange {
             Idle = 0,
             Loading,
             Prepared,
-            Paused,
-            Playing,
+            Controlled,
             Error
         };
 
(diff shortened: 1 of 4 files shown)


Source/interfaces/IStream.h
```diff
@@ -94,6 +124,9 @@ namespace Exchange {
         virtual void Callback(IStream::ICallback* callback) = 0;
         virtual state State() const = 0;
         virtual uint32_t Load(const string& configuration) = 0;
+        virtual uint32_t LastError() const = 0;
```
pwielders: Very minor, but just to avoid confusion. Lets call this Error. I do not think "Last" adds anything to the interface semantics, it might even be confusing what of you "copied" the last state into the retunr value, you get pre-empted and the error is updated to a new error, than the returned value is no longer the last error :-)

PR #225, opened 2019-10-06
Development/metadata changes

MFransen69:


Source/ocdm/open_cdm.h
```diff
@@ -187,6 +187,14 @@ EXTERNAL OpenCDMError opencdm_deinit();
  */
 EXTERNAL struct OpenCDMSystem* opencdm_create_system(const char keySystem[]);
 
+/**
+ * \brief Creates DRM system.
+ *
+ * \param system Output parameter that will contain pointer to instance of \ref OpenCDMSystem.
+ * \return Zero on success, non-zero on error.
+ */
+EXTERNAL OpenCDMError opencdm_create_system_extended(const char keySystem[], struct OpenCDMSystem** system);
```
pwielders: ????? What is this for ???

Source/ocdm/open_cdm_ext.cpp
```diff
@@ -9,9 +9,24 @@
         abort();                                                      \
     }
 
-struct OpenCDMSystem* opencdm_create_system(const char keySystem[])
+DEPRECATED struct OpenCDMSystem* opencdm_create_system(const char keySystem[])
```
pwielders: Ohh see it here, but should it not be on the header file or is this the issue that the defintion is not known there?

Source/ocdm/open_cdm_impl.h
```diff
@@ -506,12 +509,8 @@ struct OpenCDMSession {
         }
         return (false);
     }
-    inline void Metadata(string& metadata) const
```
pwielders: I thought we would refresh on load. Now we only get it during thecreation of the session. Will merge this but I guess we need a fix for this, at least on session level..

pwielders: I think I see it now on the next commit.. Oke

PR #226, opened 2019-10-10
VirtualInput: Move callback to setting to the constructor

HaseenaSainul:


Source/virtualinput/virtualinput.h
```diff
@@ -60,7 +60,7 @@ EXTERNAL void callback_touch(FNTouchEvent callback);
 
 // ================================================================================================================
 
-EXTERNAL void* virtualinput_open(const char listenerName[], const char connector[]);
+EXTERNAL void* virtualinput_open(const char listenerName[], const char connector[], FNKeyEvent keyCallback = nullptr, FNMouseEvent mouseCallback = nullptr, FNTouchEvent touchCallback = nullptr);
```
MFransen69: I'm a bit surprised that this compiles. As far as I know default parameters and nullptr are not allowed in C. Could you please check if this file can be used with a C compiler?

HaseenaSainul: Yes, you are correct, since I am checking with C++ compiler I have not noticed it. Will pass null value from the callers of these function, instead making this as nullptr

Source/virtualinput/virtualinput.cpp
```diff
@@ -128,8 +128,11 @@ namespace VirtualInput{
         Controller& operator=(const Controller&) = delete;
 
     public:
-        Controller(const string& name, const Core::NodeId& source)
+	Controller(const string& name, const Core::NodeId& source, FNKeyEvent keyCallback = nullptr, FNMouseEvent mouseCallback = nullptr, FNTouchEvent touchCallback = nullptr)
             : _channel(source, 32)
+            , _keyCallback(Core::ProxyType<Core::IIPCServer>(Core::ProxyType<KeyEventHandler>::Create(keyCallback)))
```
MFransen69: Hi Haseena, have you tested this also works for non RPI? It is just the result of a quick static code analysis so I could be wrong but I think the code will not work (crash ;) ) if you now would pass any of the pointers as 0. On the RPi they are used all three if I'm right. If you could test what happens even on a RPI when one of them is 0 (just change the RPI code calling the virtualinput_open and pass one as null) just to be sure.

HaseenaSainul: Sure I will check that

Source/virtualinput/virtualinput.cpp
```diff
@@ -130,23 +130,29 @@ namespace VirtualInput{
...
 	Controller(const string& name, const Core::NodeId& source, FNKeyEvent keyCallback = nullptr, FNMouseEvent mouseCallback = nullptr, FNTouchEvent touchCallback = nullptr)
             : _channel(source, 32)
-            , _keyCallback(Core::ProxyType<Core::IIPCServer>(Core::ProxyType<KeyEventHandler>::Create(keyCallback)))
-            , _mouseCallback(Core::ProxyType<Core::IIPCServer>(Core::ProxyType<MouseEventHandler>::Create(mouseCallback)))
-            , _touchCallback(Core::ProxyType<Core::IIPCServer>(Core::ProxyType<TouchEventHandler>::Create(touchCallback)))
         {
-            if (_keyCallback.IsValid() ==  true) {
-                _channel.CreateFactory<IVirtualInput::KeyMessage>(1);
-                _channel.Register(IVirtualInput::KeyMessage::Id(), _keyCallback);
+            if (keyCallback != nullptr) {
```
MFransen69: Only a small thing but may I propose to change the code a bit to make it more efficient and more readable (the latter might be personal preference perhaps):
Instead of the if here bring back the calls to the constructor in the initializer list like it was before but with a small change (I'll only do the _keyCallback as an example):
, _keyCallback( (keyCallback != nullptr) ? (Core::ProxyType<Core::IIPCServer>(Core::ProxyType<KeyEventHandler>::Create(keyCallback)) : (Core::ProxyType<Core::IIPCServer>(Core::ProxyType<KeyEventHandler>::Create()) )
I hope I got the brackets right, didn't try to build this ;)
Now depending if the pointer is null the correct constructor is called and not first a default constructor and later on in the body a constructor and after that an assignment operator (and the ? construction does not lead to a copy constructor by the way, this construction would also work even if ProxyType would not provide a copy c'tor or assignment operator)
Now they are initialized correctly also if the pointer is nullptr and the  if (_keyCallback.IsValid() ==  true)  below will make sure they are not registered in case the pointer was 0.

PR #228, opened 2019-10-11
[Containers] Foundations for Container plugin

mipo57:
Interface definition and functions necessary for ContainerPlugin

### DIFF
--- a/Source/processcontainers/LXCImplementation.cpp
+++ b/Source/processcontainers/LXCImplementation.cpp
@@ -6,6 +6,7 @@
 #include <lxc/lxccontainer.h>
 #include <vector>
 #include <utility>
+#include <thread>
 #include <cctype>
 
 namespace WPEFramework {
@@ -92,48 +93,6 @@ public:
 #endif
         };
 
-        class CpuInfo : public Core::JSON::Container {
-        public:
-            CpuInfo(const CpuInfo&) = delete;
-            CpuInfo& operator=(const CpuInfo&) = delete;
-
-            CpuInfo()
-                : Core::JSON::Container()
-                , Total(0)
-                , PerCPU()
-            {
-                Add(_T("total"), &Total);
-                Add(_T("percpu"), &PerCPU);
-
-            }
-            
-            ~CpuInfo() = default;
-
-            Core::JSON::DecUInt64 Total;
-            Core::JSON::ArrayType<Core::JSON::DecUInt64> PerCPU;
-        };
-
-        class MemoryInfo : public Core::JSON::Container {
-        public:
-            MemoryInfo(const MemoryInfo&) = delete;
-            MemoryInfo& operator=(const MemoryInfo&) = delete;
-
-            MemoryInfo()
-                : Core::JSON::Container()
-                , Total(0)
-                , Kernel(0)
-            {
-                Add(_T("total"), &Total);
-                Add(_T("kernel"), &Kernel);
-
-            }
-            
-            ~MemoryInfo() = default;
-
-            Core::JSON::DecUInt64 Total;
-            Core::JSON::DecUInt64 Kernel;
-        };
-
     public:
 
         LCXContainer(const LCXContainer&) = delete;
@@ -178,6 +137,8 @@ public:
             while( index.Next() == true ) {
                 _lxccontainer->set_config_item(_lxccontainer, index.Current().Key.Value().c_str(), index.Current().Value.Value().c_str());
             };
+
+            _networkInterfaces = GetNetworkInterfaces();
         }
 
         const string& Id() const override {
@@ -188,12 +149,12 @@ public:
             return _pid;
         }
 
-        string Memory() const override;
-        string Cpu() const override;
-        string Configuration() const override;
-        string Log() const override;
-        void Networks(std::vector<std::string>& networks) const override;
-        void IPs(std::vector<Core::NodeId>& ips) const override;
+        MemoryInfo Memory() const override;
+        CPUInfo Cpu() const override;
+        string ConfigPath() const override;
+        string LogPath() const override;
+        ProcessContainers::IConstStringIterator NetworkInterfaces() const override;
+        std::vector<Core::NodeId> IPs(const string& interface) const override;
         State ContainerState() const override;
         bool IsRunning() const override;
 
@@ -220,6 +181,9 @@ public:
         }
 
         private:
+            inline std::vector<string> GetNetworkInterfaces();
+
+        private:
             const string _name;
             pid_t _pid;
             string _lxcpath;
@@ -227,6 +191,7 @@ public:
             mutable Core::CriticalSection _adminLock;
             mutable uint32_t _referenceCount;
             LxcContainerType* _lxccontainer;
+            std::vector<string> _networkInterfaces;
 #ifdef __DEBUG__
             bool _attach;
 #endif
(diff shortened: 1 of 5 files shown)


Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
 namespace ProcessContainers {
 
     using IStringIterator = Core::IteratorType<std::vector<string>, const string>;
-
     struct IContainerAdministrator {
```
pwielders: Can we make this a COMRPC interface, I think that might be usefull in the near future that this information is available in "other" processes. Just inherit from virtual Core::IUnknown, you already have AddRef/Release

Source/processcontainers/LXCImplementation.cpp
```diff
@@ -90,6 +92,48 @@ class LXCContainerAdministrator : public ProcessContainers::IContainerAdministra
 #endif
         };
 
+        class CpuInfo : public Core::JSON::Container {
```
pwielders: This is a lower layer/system layer. Most likely it will be accessed, using C/C++ functions. I do not want to force these layers already to the JSON semantics (that is only applicable to JavaScript opene up functions) Prefer to keep this information as flat structs (or interfaces) this will keep the communication overhead to a minimum.
So suggest to turn this into a struct as the ProxyStub Generator does support structs and it will limit the marschalling/unmarshalling to the copy of a comaprably small piece of memory i.s.o. a long formatted string that needst to be serialized/deserialized on sending and receiving.

Source/processcontainers/ProcessContainer.h
pwielders: I think it is good to move a struct definition for MemoryInfo and CPUInfo inside this COMRPC interface. The return value of the Memory method can be this MemoryInfo struct and the return value of the Cpu can be the CPU Struct. If the user of this interface needs this information in the C++ world he does not have to "deserialize" it. Make it Lean and Mean.

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
             IContainer() = default;
             virtual ~IContainer() = default;
 
             virtual const string& Id() const = 0;
             virtual pid_t Pid() const = 0;
+            virtual string Memory() const = 0; // In bytes
+            virtual string Cpu() const = 0; // in nanoseconds of cpu time
+            virtual string Configuration() const = 0;
+            virtual string Log() const = 0;
+            virtual void Networks(std::vector<std::string>& networks) const = 0;
```
pwielders: Return an IStringIterator (it might require running over process boundaries)

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
+                ABORTING,
+                STOPPING
+            };
+
             IContainer() = default;
             virtual ~IContainer() = default;
 
             virtual const string& Id() const = 0;
             virtual pid_t Pid() const = 0;
+            virtual string Memory() const = 0; // In bytes
```
pwielders: Return the struct MemoryInfo, newly to be created with POC types? See previous remarks

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
+            virtual string Cpu() const = 0; // in nanoseconds of cpu time
```
pwielders: Return the struct CPUInfo, newly to be created with POC types? See previous remarks

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
+            virtual string Configuration() const = 0;
+            virtual string Log() const = 0;
+            virtual void Networks(std::vector<std::string>& networks) const = 0;
+            virtual void IPs(std::vector<Core::NodeId>& ips) const = 0;
```
pwielders: Also a string iterator, with the NodeId::ToHostName() ? TO be able to run over Process Boundaries?

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
+            virtual State ContainerState() const = 0;
             virtual bool IsRunning() const = 0;
+
             virtual bool Start(const string& command, IStringIterator& parameters) = 0; // returns true when started
             virtual bool Stop(const uint32_t timeout /*ms*/) = 0; // returns true when stopped, note if timeout == 0 asynchronous
 
             virtual void AddRef() const = 0;
```
pwielders: If this one inherits from IUnknown, no need to specifically specify these

Source/processcontainers/ProcessContainer.h
```diff
@@ -6,23 +6,40 @@ namespace WPEFramework {
...
             virtual uint32_t Release() const = 0;
         };
 
+        using ContainerIterator = Core::IteratorType<std::list<IContainer*>, IContainer*>;
```
pwielders: To be able to run over process boundaries, this will probably require an INterface deinitin as well.

Source/processcontainers/ProcessContainer.h
```diff
@@ -17,17 +19,30 @@ namespace ProcessContainers {
...
 
+            struct MemoryInfo {
+                uint64_t allocated; // in bytes
+                uint64_t resident; // in bytes
+                uint64_t shared; // in bytes
+            };
+
+            struct CPUInfo {
+                uint64_t total; // total usage of cpu, in nanoseconds;
+                std::vector<uint64_t> threads; // cpu usage per thread;
```
MFransen69: Consider using a std::array

Source/processcontainers/LXCImplementation.cpp
```diff
@@ -188,12 +149,12 @@ class LXCContainerAdministrator : public ProcessContainers::IContainerAdministra
...
-        string Configuration() const override;
-        string Log() const override;
-        void Networks(std::vector<std::string>& networks) const override;
-        void IPs(std::vector<Core::NodeId>& ips) const override;
+        MemoryInfo Memory() const override;
+        CPUInfo Cpu() const override;
+        string ConfigPath() const override;
+        string LogPath() const override;
+        ProcessContainers::IConstStringIterator NetworkInterfaces() const override;
+        std::vector<Core::NodeId> IPs(const string& interface) const override;
```
MFransen69: Can't we have an iterator that iterates over the interfaces and for that interface returns the name and std::array of ipaddresses?

Source/processcontainers/LXCImplementation.cpp
```diff
@@ -219,6 +180,9 @@ class LXCContainerAdministrator : public ProcessContainers::IContainerAdministra
             return retval;
         }
 
+        private:
+            inline std::vector<string> GetNetworkInterfaces();
```
MFransen69: consider using std::array (or if we use network iterator probably needs to change anyway)

Source/processcontainers/ProcessContainer.h
```diff
@@ -17,17 +19,30 @@ namespace ProcessContainers {
...
-            virtual string Cpu() const = 0; // in nanoseconds of cpu time
-            virtual string Configuration() const = 0;
-            virtual string Log() const = 0;
-            virtual void Networks(std::vector<std::string>& networks) const = 0;
-            virtual void IPs(std::vector<Core::NodeId>& ips) const = 0;
+            virtual MemoryInfo Memory() const = 0;
+            virtual CPUInfo Cpu() const = 0; 
+            virtual string ConfigPath() const = 0;
+            virtual string LogPath() const = 0;
+            virtual IConstStringIterator NetworkInterfaces() const = 0;
```
MFransen69: See my comments above where the interface is used, with hindside I better could have put them here ;)

Issue #207, opened 2019-09-13
Thunder process does not exit gracefully


goruklu:
WPEFramework process hangs when exiting. 
[wpeframework-exit-hang.txt](https://github.com/WebPlatformForEmbedded/Thunder/files/3610893/wpeframework-exit-hang.txt)

woutermeek:
Fix is merged, closing issue.

Issue #218, opened 2019-09-25
can't create a new CDM session after OpenCDM WPEProcess crashes


goruklu:
If the WPEProcess for OpenCDM crashes for any reason, it's automatically restarted but the subsequent attempts in WPE to create a new CDM session fails.

```0:01:07.464129841 23400 0xae8028a0 DEBUG      webkitmediaplayer MediaPlayerPrivateGStreamerBase.cpp:1427:cdmInstanceAttached:<play> CDM instance 0x913faac0 dispatched as context
0:01:07.466166410 23400 0xae8028a0 DEBUG          webkitopencdm CDMOpenCDM.cpp:368:Session: Creating session for 'com.microsoft.playready' keySystem and 'cenc' initDataType

-     L1 [23400]: "Creating a Session for com.microsoft.playready"
-     L1 [23400]: "Constructing the Session Client side: 0x9d570, (nil)"
-     L1 [23400]: "IPC method invokation failed for 0x10"
-     L1 [23400]: "IPC method invoke failed with error 10"
-     L1 [23400]: "Creating a Session failed. 118"
-     L1 [23400]: "Created a Session, result 0x9d570, 0"
190925-22:42:20.609447 [mod=RDKBROWSER2, lvl=INFO] [tid=23316] onConsoleLog:rdkbrowser.cpp:1072 [ConsoleAPI:18]: [1433] DRM: Ignoring initData because we have already seen it!
0:01:07.466310337 23400 0xae8028a0 ERROR          webkitopencdm CDMOpenCDM.cpp:512:operator(): could not create session id
-     L1 [23400]: "Destructed the Session Client side: 0x9d570```

Only after restarting WPEWebProcess we are able to create a new CDM session.

goruklu:
@pwielders I tried this with the latest WPEFramework code and the problem is still present.

woutermeek:
@goruklu could you please let us now if #229 solves this problem as well?

goruklu:
@woutermeek sorry for the very late response, we no longer see this issue. recent changes in #229 fixed it.

PR #229, opened 2019-10-15
Shutdown: fixes to solve issues in the shutdown sequence

HaseenaSainul:


Source/WPEFramework/PluginServer.cpp
```diff
@@ -701,14 +701,16 @@ namespace PluginHost
     {
     }
 
-	void Server::Notification(const ForwardMessage& data)
+    void Server::Notification(const ForwardMessage& data)
     {
-        _controller->ClassType<Plugin::Controller>()->Notification(data);
-		#ifdef RESTFULL_API
+        if (_controller->ClassType<Plugin::Controller>() != nullptr) {
```
pwielders: This is a bit scary. It probably indicates that there is a plugin that is not Deactivated or is not adhering to the deactivate. I would suggest placing an ASSERT here and check the callstack if this happens to correct the plugin that is issueing this notification.
If a PLugin is not properly shut down and is still firing notifications it might even be that _controller is invalid and even the call _controller->ClassType<> is failing.

Source/WPEFramework/PluginServer.cpp
```diff
@@ -701,14 +701,16 @@ namespace PluginHost
...
+            _controller->ClassType<Plugin::Controller>()->Notification(data);
+        }
+#ifdef RESTFULL_API
         string result;
         data.ToString(result);
         _controller->Notification(result);
```
pwielders: If the above situation occurs (no Controller available) this should probably also be protected.

Source/WPEFramework/PluginServer.cpp
HaseenaSainul: Yea this is happening currently we are deactivating the plugin in the Alphabetic order Z to A. So Controller plugin is get deactivated before deactivating Compositor Plugin.

Source/WPEFramework/PluginServer.cpp
HaseenaSainul: Sure will ensure this as well

Source/WPEFramework/PluginHost.cpp
```diff
@@ -53,6 +57,35 @@ namespace PluginHost {
...
+
+    private:
+        virtual uint32_t Worker() override
+        {
+            CloseDown();
+            Block();
+            return (Core::infinite);
+        }
+    };
+    ExitHandler *exitHandler = nullptr;
```
pwielders: Move all CloseDown logic in this class (including the CloseDown code). Make it the responsibility of this class to "properly" shutdown. You can Than at the end of the main, in stead of checking another variable (exitHandler), just call "Destruct" (static method). This Destruct Method can on this class checks if the internal method exitHandler is set, a closeDown already in progress (and just wait for the Thread to reach completion) and then delete it or if the Variable is not set, just call CloseDown() and wait for Completion. 
Then you can also protect this variable (exitHandler), as currently you might cause a race condition. 
What happens if a signal arrives (QUIT/TERM) and you are already at 629, so CloseDown already occured....
Lets give this class a static method called Start() that fires the thread and is the only one doing the CloseDown.
Than a second static method called Completed() that WaitsForTheThread ToCompleet

Source/WPEFramework/PluginServer.cpp
pwielders: Ohhh... Good analyses again!!! Since the Controller already has a Special status (it is a aggregated member of the server), let, by definition, not stop this plugin if we do the Close down of the server. Lets close it down after all other plugins have been stopped. This logic should go into the method where we actually "Stop"all plugins (Skip controller, by comparing it to the _controller member. and than if all plugins are deactivated deactivate this one. Than still it is a good idea to leave a DumpCallStack ere in case a plugin misbehaves .

Source/WPEFramework/PluginHost.cpp
```diff
@@ -65,46 +140,18 @@ namespace PluginHost {
             fprintf(stderr, "Signal received %d.\n", signo);
         }
 
-        if (signo == SIGTERM) {
-            if (Core::WorkerPool::IsAvailable() == true) {
-				Core::WorkerPool::Instance().Stop();
-			}
+        if ((signo == SIGTERM) || (signo == SIGQUIT)) {
+            WPEFramework::PluginHost::ExitHandler* exitHandler = new WPEFramework::PluginHost::ExitHandler();
```
pwielders: Can you move these lines 144/145 into a static method in the ExitHandler class as well (make the constructors of this ExitHandler class private so yu can not new it and only use this static Construct.
The reason for it being that you need to protect this creation. If the internal static is already "set"you do not need to start another one (so nothing to be done, a previous signal already created this). Also if the other static public method, the Destructwas already called, you also do not need to create a new exitHandler.
So we need to check for some additional conditions there.
These issues would occure if before, during or after the closing down you issue the SIGTEM/SIGQUIT signal multiple times.

HaseenaSainul: Sure, will arrange

Source/WPEFramework/PluginServer.cpp
```diff
@@ -701,14 +708,19 @@ namespace PluginHost
     {
     }
 
-	void Server::Notification(const ForwardMessage& data)
+    void Server::Notification(const ForwardMessage& data)
     {
+        if ((_controller.IsValid() == false) || (_controller->ClassType<Plugin::Controller>() == nullptr)) {
+            DumpCallStack();
+        }
+
```
pwielders: Shouldent we do an else { here ? If this situation occures, which is a "real issue", you gnerate a callstack but would not call this as this would crash the system due to nullptr access

Source/plugins/VirtualInput.h
```diff
@@ -175,7 +175,6 @@ namespace PluginHost {
                 }
 
                 ChangeIterator removed(removedKeys);
-                _parent.MapChanges(removed);
```
pwielders: I would like to investigate this one a bit further. Cause now I think we never get removal of keys mapping notifications. Why did you remove this one ?

HaseenaSainul: We are clearing the keymap table from the destructor of VirtualInput from here : https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/plugins/VirtualInput.cpp#L193
This will invoke the destructor of KeyMap: https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/plugins/VirtualInput.h#L164. VirtualInput is its parent: https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/plugins/VirtualInput.h#L159 and it is already in destruction state. Hence the line cause segfaults : https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/plugins/VirtualInput.h#L178

pwielders: As usual a very good analyses again!! Now lets think about how to solve this as now I think we are also removing notification of map changes if triggered through the PLugin (so not during destruction)

HaseenaSainul: Sure, I will try to manage it.

goruklu:
@pwielders @HaseenaSainul grea! looks like these commits will fix https://github.com/WebPlatformForEmbedded/Thunder/issues/207

Do you know if they will help with https://github.com/WebPlatformForEmbedded/Thunder/issues/218 as well ? Currently if OCDM WPEProcess crashes, that requires restarting the client (WPE Webkit) process...

HaseenaSainul:
@goruklu, yes, it will address https://github.com/WebPlatformForEmbedded/Thunder/issues/207.
I have to check https://github.com/WebPlatformForEmbedded/Thunder/issues/218, separately.

Source/plugins/VirtualInput.h
```diff
@@ -333,6 +338,12 @@ namespace PluginHost {
         virtual ~VirtualInput();
 
     public:
```
pwielders: \suggest to make this a protected method. Only derived classes now what to do with this. Also think that you can move the ClearRemovedKeys into this methods or make the ClearRemovedKeys private. Users of this class should not exercise these methods.

PR #234, opened 2019-10-22
Shutdown fix changes

HaseenaSainul:


HaseenaSainul:
There is no problem to call already closed socket, jus added check to avoid a redundant close. So removed the check.

PR #235, opened 2019-10-22
[OCDM] Recreate RPC::CommunicatorClient connection of OCDMAccessor wh…

Damian-Danylko:
…en server is interrupted by crash or deactivation

Solution for WPE-252


Source/ocdm/open_cdm_impl.h
```diff
@@ -47,17 +47,25 @@ struct OpenCDMAccessor : public OCDM::IAccessorOCDM,
...
-        _remote = _client->Open<OCDM::IAccessorOCDM>(_T("OpenCDMImplementation"));
-
-        printf("Trying to open an OCDM connection result %p\n", _remote);
+    void Reconnect() const
+    {
+        if (_client->IsOpen() == false) {
+            if (_remote != nullptr) {
+                _remote->Release();
+            }
+            _remote = _client->Open<OCDM::IAccessorOCDM>(_T("OpenCDMImplementation"));
```
pwielders: Maybe good to clean this up while we are working on this again. This is an interface that shuld have no Ext. Move all methods on Ext to the Base interface, then remove all refernces to the Ext, and use the base.
This way there is no additional round trip needed to call the second interface.

pawel-stanek-ml: +1

Damian-Danylko: I did changes on Thunder/Nanoservices, let me know if it is what you are looking for.

Damian-Danylko:
> Small remark, I am missing the removal of WPEFramework::RPC::ID_ACCESSOROCDM_EXTENSION in the ids.h file. Lets clean it all up. If you add it, it looks good to me!

Done

PR #237, opened 2019-10-25
[COMPOSITOR_CLIENT_RPI] Allow to work without Compositor Plugin on RPI

Damian-Danylko:


pwielders:
Are you sure that COMPOSITORSERVERPLUGIN this preprocessor definition is set??? I think it is never set. Did you test this ?

PR #238, opened 2019-10-28
IKeyProducer: Add: Sink class for notifications

harikris2012:
This commit adds a new Sink class in IKeyProducer for getting events
about pairing states in other plugins.

The events considered now are:
ePairingStarted
ePairingSuccess
ePairingFailed
ePairingTimedout

Signed-off-by: harikrishnan.p <harikris2012@gmail.com>

### DIFF
--- a/Source/interfaces/IKeyHandler.h
+++ b/Source/interfaces/IKeyHandler.h
@@ -71,6 +71,25 @@ namespace Exchange {
         virtual uint32_t Error() const = 0;
         virtual string MetaData() const = 0;
         virtual void Configure(const string& settings) = 0;
+
+        enum KeyProducerEvents
+        {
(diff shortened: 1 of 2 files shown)


harikris2012:
Will update the naming of events.

Can you elaborate a bit more on the interface change you suggested?

If we add a new method to IKeyHandler to pass the events(Pairing), We can get the events from implementer of IKeyProducer(Say GreenPeak) to IKeyHandler implementer(Say RemoteControl plugin). But now we will need to expose this data to other plugins, So we will need a sink class on IKeyHandler class, Right?

harikris2012:
Opened https://github.com/WebPlatformForEmbedded/Thunder/pull/242 with review comments

PR #241, opened 2019-11-02
[VIRTUALINPUT] Remove the KeyRepeatTimer watchdog thread. (Resource r…

pwielders:
…eduction)
I ran into this additional thread as this was not compiling on windows using the latest VS2019 compiler. The reference of the KeyRepeatTimer in the watchdog now needed a copy or assignment operator for a part of the template not used (VS2019 got stricter) in stead of fixing that issue (compiletime diffrentation) I chose to remove the thread as this is not needed with the WorkerPool (and timer functionality)

### DIFF
--- a/Source/plugins/VirtualInput.h
+++ b/Source/plugins/VirtualInput.h
@@ -25,8 +25,6 @@ namespace PluginHost {
             RepeatKeyTimer(const RepeatKeyTimer&) = delete;
             RepeatKeyTimer& operator=(const RepeatKeyTimer&) = delete;
 
-            static constexpr uint32_t RepeatStackSize = 64 * 1024;
-
         public:
 #ifdef __WIN32__
 #pragma warning(disable : 4355)
@@ -364,7 +362,7 @@ namespace PluginHost {
         virtual ~VirtualInput();
 
     public:
-		inline void RepeatLimit(uint16_t limit)
+	inline void RepeatLimit(const uint16_t limit)
         {
             _repeatLimit = limit;
         }
--- a/Source/plugins/VirtualInput.h
+++ b/Source/plugins/VirtualInput.h
@@ -1,46 +1,59 @@
 #ifndef __VIRTUAL_INPUT__
 #define __VIRTUAL_INPUT__
 
-#include "Module.h"
 #include "../virtualinput/IVirtualInput.h"
+
+#include "Module.h"
 
 namespace WPEFramework {
 namespace PluginHost {
 
     class EXTERNAL VirtualInput {
     private:
-        VirtualInput(const VirtualInput&) = delete;
-        VirtualInput& operator=(const VirtualInput&) = delete;
+        enum enumModifier {
+            LEFTSHIFT = 0, //  0..3  bits are for LeftShift reference counting (max 15)
+            RIGHTSHIFT = 4, //  4..7  bits are for RightShift reference counting (max 15)
+            LEFTALT = 8, //  8..11 bits are for LeftAlt reference counting (max 15)
+            RIGHTALT = 12, // 12..15 bits are for RightAlt reference counting (max 15)
+            LEFTCTRL = 16, // 16..19 bits are for LeftCtrl reference counting (max 15)
+            RIGHTCTRL = 20 // 20..23 bits are for RightCtrl reference counting (max 15)
+        };
 
-        class RepeatKeyTimer : Core::WatchDogType<RepeatKeyTimer&> {
+		class RepeatKeyTimer : public Core::IDispatch {
         private:
             RepeatKeyTimer() = delete;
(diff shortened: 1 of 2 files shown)


Source/plugins/VirtualInput.h
```diff
@@ -1,46 +1,59 @@
...
+                , _nextSlot()
+                , _job(*this)
             {
+                ASSERT(_job.IsValid() == true);
             }
 #ifdef __WIN32__
 #pragma warning(default : 4355)
 #endif
-            ~RepeatKeyTimer()
+            ~RepeatKeyTimer() override
```
MFransen69: Never thought of using it like this. Does it warn when the base class is not virtual? See a new coding guideline coming...

Source/plugins/VirtualInput.h
```diff
@@ -1,46 +1,59 @@
...
             static constexpr uint32_t RepeatStackSize = 64 * 1024;
 
         public:
 #ifdef __WIN32__
 #pragma warning(disable : 4355)
 #endif
-            RepeatKeyTimer(VirtualInput& parent)
-                : BaseClass(RepeatStackSize, _T("RepeatKeyTimer"), *this)
-                , _parent(parent)
+            RepeatKeyTimer(VirtualInput* parent)
```
MFransen69: Why changed it to a pointer? It is not allowed to pass null anyway, think & more clearly indicates that

Source/plugins/VirtualInput.h
```diff
@@ -1,46 +1,59 @@
...
 
-        class RepeatKeyTimer : Core::WatchDogType<RepeatKeyTimer&> {
+		class RepeatKeyTimer : public Core::IDispatch {
         private:
             RepeatKeyTimer() = delete;
             RepeatKeyTimer(const RepeatKeyTimer&) = delete;
             RepeatKeyTimer& operator=(const RepeatKeyTimer&) = delete;
 
-            typedef Core::WatchDogType<RepeatKeyTimer&> BaseClass;
             static constexpr uint32_t RepeatStackSize = 64 * 1024;
```
MFransen69: Think this is no longer needed

Source/plugins/VirtualInput.h
```diff
@@ -1,46 +1,59 @@
...
 #ifdef __WIN32__
 #pragma warning(default : 4355)
 #endif
-            ~RepeatKeyTimer()
+            ~RepeatKeyTimer() override
             {
             }
 
         public:
+            void DropReference()
```
MFransen69: Call it Dispose to be more consistent? Shouldn't you also call Reset() here in case the job is running?
Perhaps also add a few Asserts in other methods to check if job IsValid in case this is called at incorrect time.

Source/plugins/VirtualInput.h
```diff
@@ -339,16 +358,13 @@ namespace PluginHost {
...
         virtual ~VirtualInput();
 
     public:
-        inline void Interval(const uint16_t startTime, const uint16_t intervalTime)
-        {
-            _repeatKey.Interval(startTime, intervalTime);
-        }
-
-        inline void RepeatLimit(uint16_t limit)
+		inline void RepeatLimit(uint16_t limit)
```
MFransen69: const uint16_t limit ?

Source/plugins/VirtualInput.h
pwielders: Just obeying the rules, it should hold here as well :-)

Source/plugins/VirtualInput.h
pwielders: Because it needs to go through a template and I never ficured out how to get that corrected, without the template assuming it should use a "CopyConstructor for the real class. Hence Why I than change it to the pointer, if someone can make it work with a reference, it will be a schoolday for me.

Source/plugins/VirtualInput.h
pwielders: I asume you mean the const RepeatStackSize, Correct !

Source/plugins/VirtualInput.h
pwielders: Actually, it does not Dispose, it only Releases the reference of the internal reference to itself to avoid multiple dynamic_casts, each time the WatchDog needs to be set. So Dispose would be misleading, as the dispose will happen if the parent is destructed.

Source/plugins/VirtualInput.h
pwielders: yip. Valid point

Source/plugins/VirtualInput.h
MFransen69: Indeed commented on the wrong line

Source/plugins/VirtualInput.h
MFransen69: Good point. When I have time I'll have a look,should be doable

PR #239, opened 2019-10-30
Global: Fixes from NOS testing

rjpdasilva:
A few thunder fixes and changes drafted after NOS testing with the latest framework.

Commit 1 fixes the exported cmake files (to be used by other projects - e.g. plugins - depending on thunder), so that cmake debug related options used for thunder are not imposed on the projects importing thunder.

All of the fixes and changes are required for correct operation of Thunder under the use cases required by NOS.

PR #248, opened 2019-11-08
CompileSettings changed to PRIVATE scope

HaseenaSainul:


pwielders:
@HaseenaSainul I have no clue on CMake :-) Leave this one up to Bram to review :-)

HaseenaSainul:
Thanks @Pierre, will see comments from Bram

rjpdasilva:
Hi guys, there's one thing in https://github.com/WebPlatformForEmbedded/Thunder/pull/239/commits, where there's still a new `CompileSettingsDebug` target that other depending projects (e.g. plugins) can use if they indeed which to be "contaminated" with the options chosen for thunder.

Also, in this PR, the `CompileSettings` interface target is no longer being exported and I don't think this is what's wanted, because (besides the debug related flags we don't want to expose/contaminate), there are other settings that we do want to export, like include dirs, other cflags, etc.

rjpdasilva:
Just saw the related PR on https://github.com/WebPlatformForEmbedded/ThunderNanoServices/pull/224 and confirm that you now have to explicitly add the `CompileSettings` to all `CMakeLists.txt` file. This doesn't seem very convenient (remember there are other plugins out there) and it's not really required, if you keep exporting `CompilerSettings` as before, just without the debug related flags.

HaseenaSainul:
Thanks @bramoosterhuis for the comments, currently it is splitted to Debug and non debug, and debug settings are including only as explicitly/privately and linking wherever required. I would like to know some more clarity on this.

HaseenaSainul:
@bramoosterhuis, thanks for the detailed explanation. I will update the macros based on this.

HaseenaSainul:
@bramoosterhuis, I have moved compile defines to CompileSettings and updated as per the comments. Could you please review it

rjpdasilva:
> Ok, this is basically the same what @rjpdasilva did in https://github.com/WebPlatformForEmbedded/Thunder/pull/239/commits.
> 
> I think the problem that @rjpdasilva (correct me if I'm wrong) ran into is that compile defines like `NDEBUG` and `_DEBUG` and the options where automagically propagated to anyone that links to WPEFramework/Thunder core.

Correct.

> This is good and bad at the same time because users of core need the defines, but don't need the compile options. The defines might be used in header files so getting them automagically is ensures proper use of the headers. The bad thing is their names are super generic and will probably be used by other packages causing unwanted behavior.
> 
> So IMHO the fix is like this:
> 
> 1. Put the debug related compile options in the target `CompileSettingsDebug`; Everyone that wants to build with the same debug compile options as Thunder need to link explicitly and privately to  `CompileSettingsDebug`
> 2. Leave the compile defines in the public target `CompileSettings`, but rename the defines to be more private/explicit to Thunder/WPEFramework so they do not interfere with other headers other packages. This way we do not bother a user to have the knowledge of needing defines to use our headers in a proper way.

That's it, and this is **exactly** what I've tried to achieve through https://github.com/WebPlatformForEmbedded/Thunder/pull/239/commits and that's why there I've kept the `CompileSettings` defined and exported, just **without** the debug related stuff, which I have moved into `CompileSettingsDebug`, thus allowing other depending targets to use them anyway in case they **explicitly** choose to. With this, there's also the huge advantage of not having to change the cmake files of any of the depending targets.

rjpdasilva:
I also don't quite understand why not just use the `CMAKE_BUILD_TYPE` (maybe customizing it, extending it) instead of creating a custom `BUILD_TYPE`?

HaseenaSainul:
> I also don't quite understand why not just use the `CMAKE_BUILD_TYPE` (maybe customizing it, extending it) instead of creating a custom `BUILD_TYPE`?

Possible values for CMAKE_BUILD_TYPE are empty, Debug, Release, RelWithDebInfo and MinSizeRel. We need Debug with Optimized version, so used that.

rjpdasilva:
Hi guys, will this require any changes in cmake files for the other targets (eg plugins) that link with thunder in order to avoid the debug flags contamination?

HaseenaSainul:
No for debug defines, but if you need any debug compile options, please find config of CompleSettingsDebug  and link with CompleSettingsDebug:: CompleSettingsDebug

rjpdasilva:
So what is the https://github.com/WebPlatformForEmbedded/ThunderNanoServices/pull/224/files MR for? Isn't it still forcing the nano services to get the same debug settings as thunder?

HaseenaSainul:
Yes, we need to enable compiler debug options for plugins(to get a symbols enabled version), so that added it.

rjpdasilva:
I understand it, just that I thing there's a valid use case where you might want thunder with debug and nano services not, and with this they get "tied" together to the same config with respect to debug flags.
But let me run some tests to check.

HaseenaSainul:
Ok sure, thanks

rjpdasilva:
As for the contamination that was happening from forcing CMAKE_BUILD_TYPE, I can confirm it's OK.

I'm still not certain about automatically exporting the _THUNDER_DEBUG / _THUNDER_NDEBUG defines though... Because with this scheme, if for example I want to enable some ASSERTS() on a specific plugin, then they will only compile in if I enable DEBUG built on Thunder. Conversely, it doesn't seem right to suddenly compile in all the ASSERT() calls etc. on all other depending targets just because Thunder thunder was built for DEBUG for some Thunder specific testing/debugging.
I think it should be possible to have a DEBUG build enabled on some plugin (with ASSERT() enabled) without having to build Thunder in DEBUG, etc.

So there's still some contamination with respect to whatever is related to _THUNDER_DEBUG / _THUNDER_NDEBUG from the framework build to other external targets linking with it.
If you think about it, this is just like all those public header files that are testing the values of those defines were prefixed with `#define _THUNDER_DEBUG` for Thunder debug guilds and with `@define _THUNDER_NDEBUG` otherwise... Question is then, why the `#ifdefs` there if they can't be controlled by the modules including those header files?

rjpdasilva:
Hi @HaseenaSainul, care to comment on this please?

HaseenaSainul:
I would like to get input from @bramoosterhuis on this

rjpdasilva:
Hi @HaseenaSainul, @bramoosterhuis, could you kindly follow up on this please?

bramoosterhuis:
@rjpdasilva We need to have the debug build flags forwarded to the Thunder plugins, because of different object sizes/added functionality in the headers, e.g. see the Core::CriticalSection. If these are not linked the memory layout of the object will be different and thus will lead to crashes.

rjpdasilva:
Hi @bramoosterhuis, understand and fully agree that if those debug flags affect structures and class declarations, then there's no way around this, although I think it could be debatable to have this associated with the generic DEBUG/NDEBUG flags.
In fact, for the `Core::CriticalSection` example you give, the flag affecting that is `CRITICAL_SECTION_LOCK_LOG`, which is not controlled by `_THUNDER_DEBUG`, so I think it's actually well isolated from the generic flags. I understand however that other cases are there which are not like this (just grepping for `__DEBUG__` in the public header files shows this).

This is actually not a big deal for us as well and now the overall rationale behind it is clear.
Thanks for your explanations!

PR #256, opened 2019-11-14
Repeat key action forward added in RPI and Nexus

HaseenaSainul:


Source/compositorclient/RPI/Implementation.cpp
```diff
@@ -571,7 +571,7 @@ int Display::Process(const uint32_t data)
         time_t timestamp = time(nullptr);
         std::function<void(SurfaceImplementation*)> action = nullptr;
         if (message.type == KEYBOARD) {
-            const IDisplay::IKeyboard::state state = ((message.keyData.type == KEY_RELEASED)? IDisplay::IKeyboard::released : IDisplay::IKeyboard::pressed);
+            const IDisplay::IKeyboard::state state = ((message.keyData.type == KEY_RELEASED)? IDisplay::IKeyboard::released : ((message.keyData.type == KEY_REPEAT)? IDisplay::IKeyboard::repeated: IDisplay::IKeyboard::pressed));
```
pwielders: I guess this is the crulpit that @bramoosterhuis was facing. @HaseenaSainul can you also check and add this fix, probably to the Nexus implementation and probably also the Westeros/Weston implementation. I do not see them in the change set and they probably suffer from the sdame.

HaseenaSainul: Yes, that is already added for Nexus. But westeros case it is different and they are just forwarding all the action to the downwards (but could not able to seen any action related to repeat in the westeros code/wayland code). Also it is using Repeat functionality :https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/compositorclient/Wayland/Implementation.h#L738 not sure we have to handle it.

pwielders:
@bramoosterhuis this will allow you to remove the hack you have in netflix.

PR #257, opened 2019-11-14
WebSocket: fix for connection estblishing issue with Firefox

HaseenaSainul:


HaseenaSainul:
Thanks @pwielders, I have updated the comments

PR #258, opened 2019-11-14
PluginHost: fix to avoid segfault while calling CallStack of non exis…

HaseenaSainul:
…iting thread

### DIFF
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -626,7 +626,12 @@ namespace PluginHost {
                     uint32_t threadId = _dispatcher->WorkerPool().ThreadId(keyPress - '0');
                     printf("\nThreadPool thread[%c,%d] callstack:\n", keyPress, threadId);
                     printf("============================================================\n");
-                    PublishCallstack(threadId);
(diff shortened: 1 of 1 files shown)


Source/WPEFramework/PluginHost.cpp
```diff
@@ -626,7 +626,12 @@ namespace PluginHost {
                     uint32_t threadId = _dispatcher->WorkerPool().ThreadId(keyPress - '0');
                     printf("\nThreadPool thread[%c,%d] callstack:\n", keyPress, threadId);
                     printf("============================================================\n");
-                    PublishCallstack(threadId);
+                    if (threadId != static_cast<uint32_t>(~0)) {
+                        PublishCallstack(threadId);
+                    } else {
+                       printf("The given Thread ID is not in a valid range, please give thread id between 0 and THREADPOOL_COUNT:%d'\n", THREADPOOL_COUNT);
```
MFransen69: I guess a small thing and personal preference but I would just print between "... 0 and %d", is easier to read and understand

HaseenaSainul: Thanks @MFransen69, I have updated

PR #261, opened 2019-11-18
JSONRPC: Move SmartClient from JSONRPCClient to JSONRPCLink as SmartLinkType

HaseenaSainul:


Source/websocket/JSONRPCLink.h
```diff
@@ -2,6 +2,10 @@
 
 #include "Module.h"
 #include "WebSocketLink.h"
+#include "../plugins/Channel.h"
```
pwielders: Hopefully these includes can go no...

Source/websocket/JSONRPCLink.h
```diff
@@ -1187,5 +1186,327 @@ namespace JSONRPC {
 
     // This is for backward compatibility. Please use the template and not the typedef below!!!
     typedef LinkType<Core::JSON::IElement> DEPRECATED Client;
+    enum JSONPluginState { //Align enum values against PluginHost::IShell:state
+        DEACTIVATED,
+        DEACTIVATION,
```
pwielders: Why are DEACTIVATION and ACTIVATION still needed ?

HaseenaSainul: I could not able to delete all enum values, I have retained DEACTIVATION since to keep the same value against IShell::state

Source/websocket/JSONRPCLink.h
```diff
@@ -1187,5 +1186,327 @@ namespace JSONRPC {
...
+                return (_connection.Get<Core::JSON::VariantContainer>(waitTime, method, object));
+        }
+        bool IsActivated()
+        {
+            return (_connection.IsActivated());
+        }
+    private:
+        virtual void StateChange()
+        {
+            if (_connection.IsActivated() == true) {
```
pwielders: Lets make it an empty implementation. We do not want to have printf's in the code as printf's could be blocked (stdout -> NULL) or might be giving annoying messages wich we do not expect.

Source/websocket/JSONRPCLink.h
HaseenaSainul: It is already removed in the latest commits

Source/websocket/JSONRPCLink.h
HaseenaSainul: Sure will remove

Source/websocket/JSONRPCLink.h
HaseenaSainul: Yes I could able to remove it and tested :)

Source/websocket/JSONRPCLink.h
```diff
@@ -1187,5 +1186,323 @@ namespace JSONRPC {
...
+                    if (info[0].State.IsSet() == true) {
+                        SetState(info[0].State.Value());
+                    }
+                }
+            }
+            void monitor_on(const Core::JSON::String& parameters, const Core::JSONRPC::Error* result)
+            {
+                if (result == nullptr) {
+                    string method = string("status@") + Client::Callsign();
+                    _monitor.Dispatch<void>(3000, method, &Connection::monitor_response, this);
```
pwielders: Sorry, missed this one. Can we make all 3000 a const in this class ? It's the communication timeout.

pwielders:
Looks good. Thanks (y)

PR #264, opened 2019-11-20
[tracing] Allow using TRACE macro outside of WPEFramework namespace

artur-gebicz-ml:
Currently when using TRACE and TRACE_GLOBAL macro outside of
WPEFramework namespace, it was required to provide and additional
using namespace. In some cases, like macro usage in a header file,
this was not desired. Now TRACE macro uses full namespace and allows
to use TRACE outside of WPEFramework namespace


sebaszm:
Fix already in...

PR #271, opened 2019-11-26
Changes to handle file/image download issues

HaseenaSainul:


Source/core/Parser.h
```diff
@@ -80,7 +80,7 @@ namespace Core {
         {
             _state |= FLUSH_LINE;
         }
-        inline void PassThrough(const uint16_t passThroughBytes)
+        inline void PassThrough(const uint64_t passThroughBytes)
```
pwielders: This on an embedded system will not fly. There is no way we can parse files that are > 4G (which is a uint32_t), so uint64_t is definitielt too big.. Is there a situation where you need to exceed 64K (probably for the image file?) Would uint32_t be sufficeient than ?

HaseenaSainul: Thanks Pierre pointing this. I just increased to max value, I will check with uint32_t

pwielders: :+1:

HaseenaSainul: Changed data type to uint32_t, since its fine :)

PR #272, opened 2019-11-28
RemoteConnection cleanup while requesting deactivation from the Process

HaseenaSainul:


Source/core/IPCChannel.h
```diff
@@ -358,6 +358,7 @@ namespace Core {
             _adminLock.Lock();
 
             InternalCleanup();
+            CloseClients();
```
pwielders: The unregistration of the handlers can probably stay where it was.. They will automatically be cleared if the Clients are first Closed.
So suggest to only do the CloseClients here in the destructor and no InternalCleanup in the destructor.
Due to the closing of all "pending clients" the handlers will be unreferenced and cleared automatically

Source/com/Communicator.h
```diff
@@ -808,6 +808,7 @@ namespace RPC {
                 while (_connections.size() > 0) {
                     TRACE_L1("Forcefully closing open RPC Server connection: %d", _connections.begin()->second->Id());
                     _connections.begin()->second->Terminate();
+                    _connections.begin()->second->Release();
```
pwielders: Good catch :+1: We need this fix.

Source/core/IPCChannel.h
HaseenaSainul: We are doing the cleanup from here :https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/com/Communicator.cpp#L351 and if the handlers are not get cleared before the https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/com/Communicator.cpp#L352, it will be unregistering the handlers from here: https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/com/Communicator.h#L1087-L1088 and it is erased from IPCServerChannelType @https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/core/IPCChannel.h#L313. So in that case, https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/core/IPCChannel.h#L403 will be empty and will not get unregister from https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/core/IPCConnector.h#L767. It will assert https://github.com/WebPlatformForEmbedded/Thunder/blob/development/WPE-395/Source/core/IPCConnector.h#L742

MFransen69: I think Pierre thought (but I'm guessing now ;) ) that the code we were looking at above was the destructor and not the Cleanup method it actually is.  I think his statement is that it does not make sense to call first InternalCleanup and then CloseClients after your change to remove the handlers also from the CloseClients call. Now we would walk the client list twice (1 -> InternalCleanup,2->CloseClients ) while if just calling CloseClients would achieve the same effect (the clients that were closed but not yet cleaned up will be dealt with in the 2nd while loop in CloseClients anyway).
So I think we should also remove the unnecessary call to InternalCleanup in the CleanUp() method like you did in the destructor

I understand the problem you are trying to solve, we cannot cope with handlers being unregistered while there are still clients using them. Refcount wise it is okay, we do not leak but indeed we expect a symetric register and unregister in IPCFactory, therefore the assert.

I think your changes indeed now prevent the assert and it is not less performant (slightly better even as compared to before we do not have the while(clients) loop three times but twice while not empty ;) ). For now please leave it like this but we could now also have an assert in  ~IPCChannelServerType that the _clients.size() == 0 (and I would then call CleanUp Dispose to better indicate its function (previously it only cleaned the closed connections, now it cleans up all connections))

Source/core/IPCChannel.h
```diff
@@ -394,22 +395,25 @@ namespace Core {
             return (command);
         }
 
+        inline void UnregisterHandlers(typename ClientMap::iterator client)
```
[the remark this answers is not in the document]
HaseenaSainul: Its is an iterator(pointer), so can't make it as reference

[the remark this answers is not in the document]
MFransen69: That has of course nothing to do with the fact it is an iterator (which is not a pointer for a map). But UnregisterHandlers is called with an r-reference in the IPCChannel code which you cannot assign to a non const & (I guess you did not make it a const reference). If you make it a const reference it will work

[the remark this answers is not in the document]
HaseenaSainul: Thanks @MFransen69, I will check this

PR #273, opened 2019-11-28
IKeyHandler: interface added to handle clearing of the states/objects

HaseenaSainul:


pwielders:
This does not look like it is needed. These interfaces are refernce counted and we need to make sure that the reference counting is exercised properly, so requested interfaces need to be released :-)

PR #274, opened 2019-11-29
Added Callsign() method to IRemoteConnection interface.

neeraj-deshpande:


Source/com/Communicator.h
```diff
@@ -340,6 +341,11 @@ namespace RPC {
                     _channel->Source().Close(0);
                 }
             }
+            virtual string Callsign() const override
```
MFransen69: Wasn't so fond of the default implementation that we needed here, it indicates a problem in the structure. If you register for the IRemoteConncections you might indeed get more than you want to monitor (not all RemoteConnections are procersses in the end, therefore you need this default implementation). While thinking it through we think it is better to add an aditional (so new) interface called IProcess that has getters fot Callsign and Parent (so the latter one moves from IRemoteConnection to IProcess). Only Derivatives of RemoteProcess that actualy are monitorable (so actually have a process that can be termninated like LocalRemoteProcess and ContainerRemoteProcess then implement this IProcess interface (hint you could add this as a new base class). The users of IRemoteProcess can then do a QueryInterface and see if they can get an IProcess interface. If they do it is monitorable and then can also use that interface to getv the Callsign. I hope it is a bit clear, it is indeed a bit more compex ;), otherwise let me know and we can discuss via skype

Source/com/Communicator.h
```diff
@@ -278,8 +278,8 @@ namespace RPC {
         enum { ID = ID_COMPROCESS };
 
         virtual ~IProcess() {}
-        virtual void SetCallsign(const string& callsign) = 0;
-        virtual const string& Callsign() const = 0;
+        virtual void SetCallsign(const string callsign) = 0;
```
pwielders: Can we make it Symetrical again. or do GetCallSign/SetCallsign (not prefered) or just do allsign where the signature difference depicts the meaning. 
virtual void Callsign(const string callsign) = 0 (SETTER)
virtual const string Callsign() const = 0 (GETTER)

neeraj-deshpande: I have corrected names IProcess interface methods as suggested.

PR #280, opened 2019-12-08
[GPURAM] Move abstractions to a platform specific class.

pwielders:
- Moving the specific platform code to a single (local) platform abstraction class.
- Avoid the parsing of the TotalGPURam on every call. Do not think that runtime the number of total GPU memory bytes changes
- Avoid continous calculation of the 64 bits (on typically 32 bit platforms) TotalMemory if the freeMemory is requested.
- Avoid continous extraction of Nexus Platform configuration on every FreeRam update call.
- Avoid Hardware Platform specific preprocessor definitions in Header files, try to minimize this to OS preprocessor definitions.


Source/core/SystemInfo.cpp
```diff
@@ -24,30 +24,197 @@
...
+#if NEXUS_MEMC1_GRAPHICS_HEAP
+        if (_platformConfig.heap[NEXUS_MEMC1_GRAPHICS_HEAP]) {
+            rc = NEXUS_Heap_GetStatus(_platformConfig.heap[NEXUS_MEMC1_GRAPHICS_HEAP], &status);
+            if (rc == NEXUS_SUCCESS) {
+                freeRam += static_cast<uint64_t>(status.free);
+            }
+        }
+#endif
+#if NEXUS_MEMC2_GRAPHICS_HEAP
+        if (_platformConfig.heap[NEXUS_MEMC2_GRAPHICS_HEAP]) {
```
pwielders: @HaseenaSainul I think, but I think it is good if you test this, that the if could be an assert as the preprocessor already indicates the abscene or availability of this GRAPHICS_HEAP.

Source/core/SystemInfo.cpp
```diff
@@ -24,30 +24,197 @@
...
+            rc = NEXUS_Heap_GetStatus(_platformConfig.heap[NEXUS_MEMC2_GRAPHICS_HEAP], &status);
+            if (rc == NEXUS_SUCCESS) {
+                freeRam  += static_cast<uint64_t>(status.free);
+            }
+        }
+#endif
+    }
+
+    uint64_t TotalGPURam() const
```
pwielders: As the TotalGPURam does not change during runtime, I think it is worth the effort to move this to a specific method. This way avoiding the constant "recalcualtion" of the TotalGPU memory that is used, during the calculation of the FreeGPU. 
Also optimized the way the TotalGPUMemeory is calucalted now (only one, during the construction of the SystemInfo object.
I think, if you look at the working area of this software, it is worth the code size increase as the FreeGPU call will be issued a lot, if GPU performance measurements are used and thus will consume a lot of wasted uint64 bit calculations of the TotalGPU CPU cycles.

Source/core/SystemInfo.cpp
HaseenaSainul: Thanks @pwielders. But it seems heap handle for these can be zero as well. In arris a0 case, I found the heap handle only available for MEMC1_GRAPHICS_HEAP, the others are always zero.

PR #290, opened 2019-12-16
[ProcessContainers] C API for process contianers

mipo57:
Introduction of C-API interface for containers


Source/processcontainers/ProcessContainer.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace ProcessContainers {
+
+    IContainer* IContainerAdministrator::Find(const string& id)
+    {
+        auto iterator = Containers();
+        IContainer* result = nullptr;
+
+        while(iterator.Next() == true) {
+            if (iterator.Current()->Id() == id) {
+                result = &(*iterator.Current());
```
MFransen69: If we found what we are looking for I guess no need to continue looking ;)

Source/processcontainers/implementations/CProxyImplementation/CImplementation.cpp
```diff
@@ -0,0 +1,233 @@
...
+        result.allocated = memory.allocated;
+        result.resident = memory.resident;
+        result.shared = memory.shared;
+
+        return result;
+    }
+
+    CContainer::CPUInfo CContainer::Cpu() const
+    {
+        unsigned int cores = std::thread::hardware_concurrency();
```
MFransen69: I assume the number is also available from Thunder Core. That should be the source of OS/HW dependent info

PR #291, opened 2019-12-16
Development/bram/compositorclient closure

bramoosterhuis:
This solves the warning of a stuck callback seen on some nexus platforms:

``` 
*** 01:04:08.004 nexus_platform_client: stuck callback 3818946 msec, module priority 2 callback:0xb346ee04(0xf10300,0) interface:0x1bf1938
```
It was preventing clients to shut down properly when an EGL surface was requested. Resulting in Thunder to kill the WPEProcess. 

This fix is needed to continue NTS.

### DIFF
--- a/Source/compositorclient/Nexus/Implementation.cpp
+++ b/Source/compositorclient/Nexus/Implementation.cpp
@@ -188,11 +188,12 @@ namespace Nexus {
 
            NEXUS_Error rc = NxClient_Join(&joinSettings);
            BDBG_ASSERT(!rc);
+
+           NxClient_UnregisterAcknowledgeStandby(NxClient_RegisterAcknowledgeStandby());
 #else
            NEXUS_Error rc = NEXUS_Platform_Join();
            BDBG_ASSERT(!rc);
-#endif
-           NxClient_UnregisterAcknowledgeStandby(NxClient_RegisterAcknowledgeStandby());
+#endif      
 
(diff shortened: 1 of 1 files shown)


Source/compositorclient/Nexus/Implementation.cpp
```diff
@@ -118,6 +118,7 @@ namespace Nexus {
             virtual uint32_t Release() const override
             {
                 if (--_refcount == 0) {
+                    NXPL_DestroyNativeWindow(this->Native());
```
pwielders: Can we move this to the destructor of the SurfaceImplementation. This to avoid confusion for maintainability and to make the interface symetrical, the NXPL_CreateNativeFunction is in the Constructor, so the NXPL_DestroyNativeWindow belongs in the Destructor.

Source/compositorclient/Nexus/Implementation.cpp
```diff
@@ -223,6 +224,9 @@ namespace Nexus {
 
         virtual ~Display()
         {
+            for (auto surface : _surfaces) {
+                 NXPL_DestroyNativeWindow(surface->Native());
```
pwielders: Suggest to move this call to the SurfaceImplementaton class and set the NativeWindow ID to somthing indicating it is destructed o it is ony destructed once. Now the Release, wich moves this to the destructor, will call NXPL_DestroyNativeWindow for the seconds time.

PR #296, opened 2019-12-20
Compositorclient [RPI/Wayland]: Add/release based on the instance count and ensure releasing after use

HaseenaSainul:


pwielders:
With the new COMRPC generations and the changes of the WPEComposition in the seperate AppEngines, this issue is tackled.

PR #297, opened 2019-12-20
Clients cleanup while closing the connection as well

HaseenaSainul:


pwielders:
Fixed by PullRequest 350

PR #305, opened 2020-01-10
Remove floats and double from a Thunder Framework code

Damian-Danylko:


Source/core/SystemInfo.cpp
```diff
@@ -268,15 +268,14 @@ namespace Core {
...
             for (int i = 0; i < numFields && i < 10; ++i) {
                 CurrentTickCount += CpuFields[i];
             }
-
-            double DeltaTickCount = CurrentTickCount - previousTickCount;
-            double DeltaIdleTime = CurrentIdleTime - previousIdleTime;
+            uint64_t DeltaTickCount = CurrentTickCount - previousTickCount;
+            uint64_t DeltaIdleTime = CurrentIdleTime - previousIdleTime;
 
             SystemInfo::m_cpuload = ((DeltaTickCount - DeltaIdleTime) / DeltaTickCount) * 100;
```
MFransen69: Do we still get the same result as before? If I'm right, but I could be easily wrong ;), the base type will remain uint64_t while doing the division, so you will lose the remainder and the *100 will always add two zero's, so the percentage calculated is now not as expected.

Damian-Danylko: True, I will provide a valid solution.

PR #310, opened 2020-01-17
[CyclicBuffer] Fixed the deadlock situation during the timed lock()-u…

anjalirajan:
…nlock() sequence.

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -87,6 +87,7 @@ namespace Core {
     uint32_t CyclicBuffer::SignalLock(const uint32_t waitTime)
     {
 
+        AdminLock();
         uint32_t result = waitTime;
 
         if (waitTime != Core::infinite) {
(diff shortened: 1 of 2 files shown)


pwielders:
Locking the lock with an overall lock seems scary. Is there a test case that shows this issue and that can be run to show that with this fix it is rsolved ?

anjalirajan:
Please have a look on the test https://github.com/WebPlatformForEmbedded/Thunder/blob/anjali/unittest/Tests/core/test_cyclicbuffer.cpp#L215

Without the fix https://github.com/WebPlatformForEmbedded/Thunder/pull/310/commits/b0bfa074c8844f8b1c485b79565fec0630a0cf6a#diff-ccdd19b6cdb0b67eb9493ea7128fc51cL160, lock_unlock test was hanged .

From my understanding(https://stackoverflow.com/questions/16522858/understanding-of-pthread-cond-wait-and-pthread-cond-signal/16524148), conditional variables are commonly used inside the pthread_mutex_lock() & pthread_mutex_unlock(). ie. why added https://github.com/WebPlatformForEmbedded/Thunder/pull/310/commits/b0bfa074c8844f8b1c485b79565fec0630a0cf6a#diff-ccdd19b6cdb0b67eb9493ea7128fc51cR90.

PR #314, opened 2020-01-24
Generated ProxyStubs_Communicator.cpp from Communicator.h using StubGenerator.py.

neeraj-deshpande:
Latest  StubGenerator.py has now support for @interface tag as well as corrected behavior for @stubgen:skip tag.


Source/com/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(TARGET ${NAMESPACE}COM)
 add_library(${TARGET} SHARED
         Administrator.cpp
         Communicator.cpp
+        ProxyStubs_Communicator.cppp
```
pwielders: New definition for a CPP file ?

Source/com/ProxyStubs_Communicator.cpp
```diff
@@ -0,0 +1,391 @@
+//
```
pwielders: Lets not checin generated files, they lead to confusion and potential issues!

neeraj-deshpande: I have modified to generate ProxyStubs_Communicator.cpp during build process.

PR #320, opened 2020-01-29
[VIRTUALINPUT] Extend the IPC VirtualInput to give access to the open…

pwielders:
… channels.


Source/plugins/VirtualInput.h
```diff
@@ -691,19 +694,106 @@ namespace PluginHost {
...
+            Iterator()
+                : _server(nullptr)
+                , _index(~0)
+                , _element()
+            {
+            }
+            Iterator(VirtualInputChannelServer& server)
+                : _server(&server)
+                , _index(~0)
+                , _element()
```
pawel-stanek-ml: Could be marked explicit and use delegating ctor i..e
explicit Iterator(VirtualInputChannelServer( ) : VirtualInput() {}

Source/plugins/VirtualInput.h
```diff
@@ -691,19 +694,106 @@ namespace PluginHost {
...
+            }
+
+            Iterator& operator=(const Iterator& rhs)
+            {
+                _server = rhs._server;
+                _index = ~0;
+
+                if (_element.IsValid() == true) {
+                    _element.Release();
+                }
```
pawel-stanek-ml: This is Reset() so Reset() could be called here instead of duplication.. But why copying or assigning from another Iterator makes this to point at the beginning of the "iterable" i.e. element & index are not coipied?

pwielders: Will update to use Reset. Like to avoid Code duplication. But given you second remark I think it is better to actually copy the values ;-)

Source/plugins/VirtualInput.h
pwielders: Agree, good practice to use exlicit on single argument constructors. Will adapt. I guess with the delegation you mean explicit Iterator(VirtualInputChannelServer&) : Iterator() {} ? Will add that one as well. Need to get used to constructor delegation, in my days in was not available ;-)

pwielders: The delegation constructor, agree if it is a large overlapping set, but here it would only benefit for 2 variables and 1 variable would be set first to null and during cunstruction time be changed to the server. 
Now it becomes more of a trade-off will the compiler optimize away the setting of _server to null and then to server paremeter or will it set this variable now twice? 
I do think for maintainabilty it is good to have Constructor delegation, so I will still go for Constructor forwarding, but will also check what the compiler does.....

pawel-stanek-ml: As discussed offline - yes. It's my bad. I missed the _server needs to be initialized with the passed in arg :D. This defeats the purpose as delegating ctors require no other initialization to be done in the initialization list. Feel free to ignore.

PR #322, opened 2020-01-30
[Threading] WorkerPool Revoke() now guarantees job is not running anymore

mipo57:
Forces Revoke() method of workerpool to wait in case the job is running at the moment. Fixes potential problems with hard to reproduce timing-based bugs

### DIFF
--- a/Source/com/Administrator.h
+++ b/Source/com/Administrator.h
@@ -248,7 +248,7 @@ namespace RPC {
         {
             return (_factory.Element());
         }
-        void Clear()
+        void Dispose()
         {
             _message.Release();
             _channel.Release();
--- a/Source/core/WorkerPool.cpp
+++ b/Source/core/WorkerPool.cpp
@@ -16,12 +16,16 @@ namespace WPEFramework {
 			_metadata.Slots = threadCount;
 			_metadata.Slot = counters;
 			_instance = this;
+
+			_workerStatuses = new WorkerStatus[threadCount];
 		}
 
 		WorkerPool ::~WorkerPool()
 		{
 			_handleQueue.Disable();
 			_instance = nullptr;
+
+			delete[] _workerStatuses;
 		}
 	}
 }
--- a/Source/core/WorkerPool.h
+++ b/Source/core/WorkerPool.h
@@ -62,6 +62,9 @@ namespace Core {
                 _job->Dispatch();
                 _job.Release();
             }
+            inline uint64_t Id() {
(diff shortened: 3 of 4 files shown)


Source/com/Administrator.h
```diff
@@ -248,7 +248,7 @@ namespace RPC {
         {
             return (_factory.Element());
         }
-        void Clear()
```
pwielders: Do not rename this. Whenever such an object is part of a workerpool, the Workerpool will "look" for the method Clear (using Type Traits) if the method is found, it will call it before the object is returned to the pool. If you rename it, it will not find the method and the object will not be cleared.

Source/core/WorkerPool.cpp
```diff
@@ -16,12 +16,16 @@ namespace WPEFramework {
 			_metadata.Slots = threadCount;
 			_metadata.Slot = counters;
 			_instance = this;
+
+			_workerStatuses = new WorkerStatus[threadCount];
```
pwielders: new always scares me. It is expensive (relatively) and I think we can solve the issue ar hand, without any new objects. You can take for granted that a Job always has a unique Id, the Address of the IDispatch interface. It will always be unique,. So I think, you can "check" if a Job is running by simply asking each minion if the IDispatch address it is currently executing is equal to the IDispatch of the Job that you would like to revoke. If that is the case, you found the minion that is executing the Job you would like to "Remove". 
Next step is to add a functionality to that Minion to signal another thread that the Job is completed.
There you have 2 options, busy waiting, not preferred as a Job execution is undeterministic, or waiting for an Event.
The latter could be achieved by simply "adding an event to each miniom, and signal a waiting method in the minion for a designated IDispatch. So I think we do not need an additional calss and we can prepare all at creation of the minion.

Source/core/WorkerPool.h
```diff
@@ -244,6 +306,7 @@ namespace Core {
             Job newRequest;
 
             while ((Running() == true) && (_handleQueue.Extract(newRequest, Core::infinite) == true)) {
+                _workerStatuses[index].JobStarted(newRequest.Id());
```
pwielders: I guess this would even by a race condition, what if the Job is extracted, a request to Revoke the Job comes in, but before the _workerStatuses is updated, (preemption between Extract and JobStarted) the whole checking sequence was done. It will assume that the job was *NOT* running and thus return. However is the minion is scheduled to run again, he will assume the job is safe to run.

PR #325, opened 2020-01-31
Modified compositorclient to remove surface from _surfaces list when the surface is unregistered.

neeraj-deshpande:


Source/compositorclient/RPI/Implementation.cpp
```diff
@@ -651,9 +650,13 @@ inline void Display::Unregister(Display::SurfaceImplementation* surface)
 
     _adminLock.Lock();
 
-    std::list<SurfaceImplementation*>::iterator index(
-        std::find(_surfaces.begin(), _surfaces.end(), surface));
-    ASSERT(index != _surfaces.end());
+    for (auto index = _surfaces.begin(); index != _surfaces.end();) {
```
MFransen69: Hi Neeraj, I think in principle the fix is correct. However it is a little ;) inefficient. The surface can only be in the list one, so please use a find and erase the iterator returned by the find (if !-= end of course)

PR #327, opened 2020-02-03
FindNXCLIENT: add nxserver header only if it available

HaseenaSainul:


pawel-stanek-ml:
Looks good to me. Added @artur-gebicz-ml to have a look too

cmake/modules/FindNXCLIENT.cmake
```diff
@@ -38,8 +38,12 @@ if(EXISTS "${LIBNXCLIENT_LIBRARY}")
     include(FindPackageHandleStandardArgs)
 
     set(NXCLIENT_FOUND TRUE)
+    set(LIBNXCLIENT_INCLUDE ${NXCLIENT_INCLUDE})
+
+    if(EXISTS "${NXCLIENT_NXSERVER_INCLUDE}")
```
artur-gebicz-ml: Right now you are only hiding the problem here. When the package will do find_package(nxclient) and it needs nxserverlib.h you won't get message that you are missing files. This should be done in other way. I would go with creating a second variable that is LIBNXCLIENT_NXSERVER_INCLUDE and store that information there. If then the target want to have access to nxserver then it should explicitly mark it eg:
```
target_compile_definitions(${TARGET}
    PRIVATE
        ${NXCLIENT_INCLUDE}
        ${NXCLIENT_NXSERVER_INCLUDE})
```
And the targets that don't need to use that nxserver part should be separated.
The other way around is to split find_package to 2 recipes 1 for nxclient and 2 for the nxclient_server

HaseenaSainul: Thanks @artur-gebicz-ml, since we have to check only header here, so merged both in the same cmake. will split it into two

HaseenaSainul:
This was added to support nxserverlib.h header search for NEXUS-SERVER-EXTERNAL case. But the dependency is able to handle using FindNXServer.cmake itself. So closing this PR

PR #339, opened 2020-02-17
Modified to automate proxy stub generation for ITraceIterator & ITraceController

neeraj-deshpande:


Source/protocols/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(TARGET ${NAMESPACE}Protocols)
 find_package(ZLIB REQUIRED)
 
 ProxyStubGenerator(NAMESPACE "RPC" INPUT ${CMAKE_CURRENT_SOURCE_DIR}/../com/Communicator.h)
+ProxyStubGenerator(NAMESPACE "Trace" INPUT ${CMAKE_CURRENT_SOURCE_DIR}/../com/ITracing.h)
```
pwielders: Also make these modifications in the Source/com/CMakeLists.txt files. Some deplioyment don'r combine these two libraries

pwielders:
I think we also need to remove IStringIterator.cpp and IValueIterator.cpp as these can also be generated from the Interface definitions

neeraj-deshpande:
> I think we also need to remove IStringIterator.cpp and IValueIterator.cpp as these can also be generated from the Interface definitions

Tried to create proxy-stubs for IStringIterator and IValueIterator classes in  Thunder/com using StubGeneraor script. 
However as these classes are actually typedef to IIteratorType template class.
According to Sebastian, its a bigger task to treat typedef to template class as a regular class and create proxy-stubs.

PR #345, opened 2020-02-26
BD-775 Thunder, Added license attribute in URL.cpp and URL.hfile, rem…

Sukanya673:
BD-775 Thunder, Added license attribute in URL.cpp and URL.hfile, removed some attributes from NOTICE file since it is not required


pradeeptakdas:
Hi @woutermeek , Apache licenses are added in newly modified files [URL.h & URL.cpp],  please review and merge? Thanks!

PR #351, opened 2020-03-05
[SystemInfo] Init _provisioning variable

harikris2012:
The pointer variable _provisioning was not initiallized to nullptr in
the constructor.

And since c++ doesnt init pointer data types to nullptr when an instance
of the class is created. This can result in _provisioning having
junk/invalid value.

Some spordic crashes were observed, Where framework crashes on init.
While debugging it was observed that we are trying to Release() an
invalid _provisioning handle. Which is causing the crash.


harikris2012:
@pwielders Can you please review

